PR #15, opened 2020-03-16
add ?sort=prop feature

dill0wn:
Allow results of `/countries` to be sorted by an optional query param.

ex: `/countries?sort=todayCases`

I had to make some changes to the package.json in order to run the server. I was getting build errors from node-gyp on my version of node/npm, which looks like it was a dependency of quick.db. Updating quick.db to latest beta seemed to fix the issue.

I also noticed that request was no longer being used, instead replaced by axios. So I removed it. Which made the lack of 'cors' throw an error, so I added that back in.

```
$ npm --version
6.13.4
$ node --version
v12.16.1
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+json.sqlite
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules/
 json.sqlite
+.vscode/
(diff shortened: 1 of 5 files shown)


EliteDaMyth:
will merge when i get on pc and pull on my server good work :)

EliteDaMyth:
seems good mergin.

PR #58, opened 2020-03-19
Add death rate attribute to country endpoint

ebwinters:
Just a small feature to make it more convenient to retrieve updated death rates for countries, and hopefully be able to sort on death rate once the sort feature is merged.


dicedtomatoreal:
There is no need for this to be in the API, since the API's job is to give user data. The user can do whatever they want with that data, like calc the death rate.

PR #69, opened 2020-03-20
formate update from endpoint /getAll for toLocaleString

anajuliabit:


EliteDaMyth:
A timestamp is better in this case at it can be converted on the client side as well so its not really an issue

PR #77, opened 2020-03-21
Add editorconfig and function to filter the list of countries

ThiagoNelsi:
## Filter function

>The function allows the user to insert a filter in the list of countries through the URL;
>
>
> https://corona.lmao.ninja/countries?filter={key}'{operator}'{value}
>
>
>### Example
> https://corona.lmao.ninja/countries?filter=cases'>'1000
>
> ###Allowed operators:
>-  '=' return countries with the key equals to the passed value
>         https://corona.lmao.ninja/countries?filter=todayCases'='50
>-  '>' return countries with the key higher then the passed value
            https://corona.lmao.ninja/countries?filter=cases'>'1000
>-  '<' return countries with the key lower then the passed value
            https://corona.lmao.ninja/countries?filter=casesPerOneMillion'<'500
>
>###Allowed keys:

>- cases
>- todayCases
>- deaths
>- todayDeaths
>- recovered
>- active
>- critical
>- casesPerOneMillion

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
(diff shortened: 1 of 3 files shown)


dicedtomatoreal:
This should be done on the client side.

PR #78, opened 2020-03-21
update readme for country historical data

ebwinters:


ebwinters:
ayoo merge this so people know how to use the endpoint

PR #79, opened 2020-03-22
Caching Covid-19 data to S3 JSON structures

AidHamza:
#### Description : 
Upload all the COVID-19 results to S3 structures : 
- {COUNTRY_CODE}.json
- world.json
- overall.json
- {SCRAPE_DATE}/COUNTRY_CODE.json


AidHamza:
Sorry, I was trying to PR my repo

dicedtomatoreal:
Yeah no this is not being merged. Please make a fork of the api and use S3, from my point of view requesting S3 a lot of times per second (by a lot I MEAN A LOT :joy:). This is why we use Redis since it can handle the amount of requests we get every second.

AidHamza:
@dicedtomatoreal by default Github PR to the base repo :p 
oh no, Querying S3 that much is not an option, I modified the API to be just a worker to S3, then putting Cloudfront as CDN

the consumption of Cloudfront is going to be less than 500GB per month, since its gonna be consumed from some folks trying to create a web version, and a mobile one

PR #85, opened 2020-03-22
Update historical.js

FireFeathers06:
null isnt a thing in python hence gives error please update :)

### DIFF
--- a/funcs/historical.js
+++ b/funcs/historical.js
@@ -57,7 +57,7 @@ var historical = async (keys, redis) => {
     }
     result.push({
       country: parsedCases[b][1],
-      province: parsedCases[b][0] === "" ? null : parsedCases[b][0],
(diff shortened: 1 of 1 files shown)


jessestuart:
@FireFeathers06 
Sure, but it _is_ a thing in JSON, arguably the _lingua franca_ of the web for passing data around. To that end, you *do* have a `None` type in Python that serves the save role as `null` in JS, as well as a built-in library to handle this very conversion; see, for example, this [Stack Overflow response](https://stackoverflow.com/a/27140103/12529168):

```python
>>> import json
>>> data = '''
... {
...   "abc": null,
...   "def": 9
... }
... '''
>>> json.loads(data)
{'def': 9, 'abc': None}
```

Not too bad, right?

Most importantly: `null` and `"null"` mean very different things to the javascript runtime (really, any language where you're using a JSON serializer/deserializer). Try to substitute one for the other, and you're in for a bad time. 🙃

Issue #92, opened 2020-03-22
the API getting the wrong country


WWGate:
Hello,

I'm trying to get Oman country data but the API returns data for Romania, any help please

RyanHarlow:
looks like its a bug. there is line in server.js that retrieves the country
  e => e.country.toLowerCase().includes(req.params.country.toLowerCase())

because rOMANia includes oman its returning romania first

EliteDaMyth:
@RyanHarlow that pr can bream stuff instead can you make it so if there is a request paramenter strict it will search for the exact name. So it doesn't break current projects

EliteDaMyth:
Break*

EliteDaMyth:
Fixed with merging of the pr. Will update soon

PR #93, opened 2020-03-22
fix wrong country being returned as described in issue #92

RyanHarlow:
When requesting the country Oman at the endpoint /countries/oman the api would return romainia because it was using .includes and romainia includes the string Oman.  I changed this to only return if the country is equal to the one being requested.


RyanHarlow:
wooops my bad damn let me try again

PR #98, opened 2020-03-22
Updated column count

alitas:
worldometers.info added "source" column, so the number of columns is now 8


dicedtomatoreal:
k

EliteDaMyth:
@all-contributors please add @alitas for bug

PR #103, opened 2020-03-23
Ryanharlow

RyanHarlow:
fixed wrong country name issue #92 by adding strict mode to match name exactly. Also updated the readme.md with strict mode


EliteDaMyth:
Merging :)

Issue #121, opened 2020-03-23
Countries EndPoint : countryInfo returns NO DATA for many countries


55abhilash:
Hi,

When I hit [https://corona.lmao.ninja/countries](https://corona.lmao.ninja/countries), an array of JSON objects is returned, each element of array is information for one country in JSON.
In each JSON, there is a JSON key-value pair for "countryInfo". This information is returned for most countries. However, it is returned as NO DATA for many countries, Eg. USA, UAE, Syria, etc.
![Screenshot from 2020-03-23 22-09-58](https://user-images.githubusercontent.com/9768146/77340277-1299bd80-6d53-11ea-93d3-1b73be2fbd52.png)

55abhilash:
It looks like the issue is due to difference in name of countries in the https://www.worldometers.info/coronavirus/ main table which is used by us, and the ISO information statically included here: https://github.com/NovelCOVID/API/blob/master/utils/country_utils.js.
Eg. In the worldometers.info table, country is called "USA" while in our utils file it is "United States". Due to this mismatch, the call to getCountryData returns NO DATA.
I will submit a pull request with the changes if you think this analysis is right.

buster95:
Was resolve on #117

stoutlabs:
The country names and capitalization have been changing as well... which is quickly becoming frustrating. e.g. `Korea, South` became `korea, south` and most recently it is now `s. korea`.  ISO names would be best, and always consistent.

buster95:
Country names haven't been changing, data is showed like `https://www.worldometers.info/coronavirus`
you can see here
![image](https://user-images.githubusercontent.com/15637669/77361907-b6fd0f00-6d15-11ea-91eb-1bda1231c6a0.png)

If you are talking about this endpoint `https://corona.lmao.ninja/countries/{country name}`, `country name` parameter have been changing

stoutlabs:
Ah, yes... that's what I was referring to. (Timeline data as well.)

EliteDaMyth:
already a pr trying to fix.

PR #117, opened 2020-03-23
[Countries] Synonyms or possible names for a lot of countries

buster95:
Adding synonyms for a lot of countries, homologate 
https://www.worldometers.info/coronavirus/     Country Catalog with this API

This PR has:
- [x] Pairing countries with worldometers web page
- [x] Some spanish country names for searches ex. `Korea del Sur` -> `South Korea`, `Italia` -> `Italy`
- [x] Improvements on countries/{country name} endpoint found countries easier
- [x] Code reduction

This PR
resolve #121 
resolve #127 
resolve #128
resolve #150

PR #122, opened 2020-03-23
Pull Request for #121: Changed country names in variable "countryData" 

55abhilash:
Changed country names in variable "countryData" according to names on worldometer so that getCountryData returns properly.


buster95:
I add synonyms or possibles names, this PR isn't necessary, was resolve on #117

55abhilash:
But the synonyms or possible names are not included in array "countryData". They are in a different array, which is not being accessed in "getCountryData". 
If your synonyms are to be used, kindly change the code in function "getCountryData" accordingly.

buster95:
Idea is synonyms just for searches getCountryData only return standardized country name

EliteDaMyth:
Does this have to be merged

55abhilash:
@buster95 : This is the first few lines of code of function getCountryData : 
`const getCountryData = function (countryName) {
    for (const index in countryData) {
        if (**countryData[index].country.toLowerCase() == countryName.toLowerCase()**) {
            const data = countryData[index]`
Notice the comparison in **if**.

And this is called in getCountries.js line 54:
`            let country =
                cell.children[0].data ||
                cell.children[0].children[0].data ||
                // country name with link has another level
                cell.children[0].children[0].children[0].data ||
                cell.children[0].children[0].children[0].children[0].data || "";
result.push({ country, countryInfo: countryUtils.getCountryData(**country**) });`
Here, the variable **country** is coming from worldometer html table. Value of this variable is compared to **countryData** value as discussed above. 
Hence, if you use standardized names in countryData, and not names similar to worldometer html table, the comparison in **if** will fail, and it will:
`return { iso2: 'NO DATA', iso3: 'NO DATA', _id: 'NO DATA', lat: 0, long: 0, flag:`

I hope this explanation is verbose enough :P

buster95:
ok 
`{ country: 'United States', iso2: 'US', iso3: 'USA', _id: 840, lat: 38, long: -97, possibleNames: ['USA', 'Estados Unidos'] },`

you can look for `USA`, `Estados Unidos` or `United States`
ex.
**`countries/usa`**
**`countries/estados%20unidos`**
**`countries/united%20states`**

all urls response with the same data, displayed country name will be **`United States`** because is on country field.

you can do unit tests for `getCountryData` function to understand how it work

55abhilash:
Problem is not while doing **/countries/usa**, it is during **/countries**. Country name in such case is United States and worldometer names is USA. That's the mismatch.
I can show you the error on discord. Is it possible?

EliteDaMyth:
we standardise the country name. the names are not the same as worldometers

55abhilash:
Ok. In that case, `{ if (countryData[index].country.toLowerCase() == countryName.toLowerCase())`
this should compare countryName (From WorldoMeter) to isoCountries array, which contains all possible country names and synonyms by @buster95 instead of countryData, which contain only standardised names. In this way, we standardise the country names, and data is also fetched for countries with non standard names on worldometer.

Issue #133, opened 2020-03-24
https://corona.lmao.ninja/jhucsse Data In wrong places


SaadAsif1:
While looking at the https://corona.lmao.ninja/jhucsse api most of the data are in the wrong places such as "country" is showing the province, "province" is showing a number, "updatedAt" is showing a state, and more.

buster95:
Wrong fields 🤨
![image](https://user-images.githubusercontent.com/15637669/77382702-a0bb7700-6d46-11ea-98bc-fad22b9c78ba.png)

RyanHarlow:
working on it

SaadAsif1:
Let us know when fixed thank you.

RyanHarlow:
@SaadAsif1   i submitted the pull request, just needs to merged and deployed

SaadAsif1:
@RyanHarlow Thanks

PR #134, opened 2020-03-24
fix /jhuLocations endpoint json was scrambled where data was in the wrong place issue #133

RyanHarlow:
fixing issue #133


EliteDaMyth:
@all-contributors please add @RyanHarlow for bug

PR #141, opened 2020-03-24
fix for worldometers removing recovered column, active cases is now i…

JamesShelley:
Fix for worldometers removing recovered column, active cases is now in column index 5.


EliteDaMyth:
I just fixed this like 2 seconds before your PR. Thanks for tryna help though. :)
https://github.com/NovelCOVID/API/commit/930a8a0e55849b90e763ed4d0a7e5ac8f50fd551

JamesShelley:
No worries!

PR #162, opened 2020-03-24
[RFC / Proposal]: Begin adding test cases for core API.

jessestuart:
### What does this PR do?
I'd like to suggest we begin introducing some basic test cases around the core API. This is both a way to improve API stability and can be a great way for less experienced developers / new contributors to chip in and get familiar with the codebase without the pressure of touching critical-path code. If adopted, as test coverage improves, this approach allows both contributors and reviewers to iterate more quickly **and with confidence** that a given change is both correct (i.e., it returns what it says it should) and **isolated** (i.e., it doesn't unexpectedly break some other code path).

I've also set up a super simple CI pipeline on CircleCI (who I've used for years; their free plan for open-source projects is amazing). Currently, it just checks out the repo, installs dependencies, and verifies that tests pass. (See an example run [here](https://circleci.com/gh/jessestuart/API/18))

### Some notes on node/javascript testing best practices
To ensure that we can expect constant outputs whenever you run the tests, it's often necessary to mock out external dependencies — we're trying to isolate the **code itself as the unit under test**, so given the same input, it should return the same output. In this case, that means:
- Mocking out Redis (we can assume our cache works, and it screw things up anyway)
- Intercepting HTTP calls to worldometer and substituting the response data with a "fixture", i.e., a snapshot of the page at some point in time. In our case, just a giant string of HTML that I pulled from the page. **Yes**, this won't catch bugs caused by them changing the format on their end, but I'd argue 1) that should be a different test (integration test), and 2) ideally our code could be robust enough to handle multiple possible formats, especially when it comes to scraping. We have no ideas whether a given UI change is permanent, or part of an A/B test being run and thus temporary, or even an inadvertent regression. By keeping fixtures-as-code, we have snapshots of how the page was structured at a given point in time, and try to implement parsers that can handle several (which can then be easily verified just by running the tests).

Let me know what y'all think!


ebwinters:
@EliteDaMyth Why did you close this

jessestuart:
@EliteDaMyth Any response to @ebwinters here? I noticed you also closed another one of my PR's (#81) without comment.

It's cool if you don't believe this contribution is appropriate for the project at this time, but it's expected in the OSS community to at least leave a comment explaining your thinking, or better yet, suggesting what changes could be made to get a diff merged. Closing PR's just to "clear out the queue" doesn't send a super friendly message to contributors, FWIW 🙂

ebwinters:
looks like you put a lot of time int this PR. I think the PR should remain open so people can look at it, but not sure it can be merged rn anyways. Once our infrastructure is fixed to run on load balancer we will be using GH actions, not circleCI

PR #178, opened 2020-03-25
Update countryData in country_utils.js

moinahmed9999:
Updated country names in countryData so that it matches with information(country names) according to https://www.worldometers.info/coronavirus/ . 
This way the api will properly show countryInfo{lat, lang, iso and flag urls} which is was not showing earlier because mismatch in country names.


moinahmed9999:
Thank you !! 
How long will it take to update in the api itself as i still see countries with isos NO DATA and lat, long as 0

EliteDaMyth:
doing in 10-20 mins

moinahmed9999:
> doing in 10-20 mins

Alright !!

PR #188, opened 2020-03-25
Updated to work with getYesterday's PR

### LINKED ISSUE #181: Possible Suggestion - Adding Yesterday's Statistics
I was messing about with some of the original API code from when you first released it publicly,
I managed to get a scraper working for the 'yesterdays statistics' simply by doing:

```js
 let response;
    try {
      response = await axios.get("https://www.worldometers.info/coronavirus/#nav-yesterday");
      if (response.status !== 200) {
        console.log("Error", response.status);
      }
    } catch (err) {
      return null;
    }
  
    // to store parsed data
    const result = [];
  
    // get HTML and parse death rates
    const html = cheerio.load(response.data);
    const countriesTable = html("table#main_table_countries_yesterday");
    const countriesTableCells = countriesTable
      .children("tbody")
      .children("tr")
      .children("td");
  
     // NOTE: this will change when table format change in website
     const totalColumns = 10;
     const countryColIndex = 0;
     const casesColIndex = 1;
     const todayCasesColIndex = 2;
     const deathsColIndex = 3;
     const todayDeathsColIndex = 4;
     const curedColIndex = 5;
     const activeColIndex = 6;
     const criticalColIndex = 7;
     const casesPerOneMillionColIndex = 8;
     const deathsPerOneMillionColIndex = 9;
```

I think this should be added to the API, unless the JHU endpoint does the same functionality.
Cerberus.
[Full Scraper Code](https://github.com/NovelCOVID/API/files/4381369/message.txt)

Mr-Cerberus:
Modified the scraper.js to work with the [getYesterday.js PR Integration](https://github.com/NovelCOVID/API/pull/183#issuecomment-603947258)
that way you can put this function to work. closes #181


ebwinters:
I think you should also an endpoint to this PR in `server.js` so when we merge it is ready to go. This should fall under v2 so make the route `v2/yesterday` or something like that

EliteDaMyth:
> I think you should also add an endpoint to this PR in `server.js` so when we merge it is ready to go. This should fall under v2 so make the route `v2/yesterday` or something like that

seems stale. closing. If you are able to make the changes do open another PR!

ebwinters:
Reviewing today

ebwinters:
Don't merge this yet

PR #198, opened 2020-03-26
Improve the parsing of the table data from WorldMeteres

ben-richardson-moneysmart:
Update `casesPerOneMillion` and `deathsPerOneMillion` to handle comma grouped values and empty values in the WorldMeteres table.

Fixes the issue where values over 1000 are shown just as the first digit.
Fixes issue where value comes back as `null` if the column was empty. 0 seems like better value than a null + consistency with the other int columns of the API


ebwinters:
closes #204

PR #229, opened 2020-03-27
[Feature] Introducing Searching for ISO 3166 and Countries Ids

buster95:
This PR add the possiblity to do searches with ISO and Id
ex.
/countries/{iso2}
/countries/{iso3}
/countries/{country id}
/countries/{country name}

resolve #226
closed #228

PR #230, opened 2020-03-27
Allow Searching for ISO codes

Dannyps:
This will allow the `:country` key in the `/countries/` endpoint to be either the (partial) name or an iso2 or iso3 code. No further manual fixes required for these :)

Please review and leave a note if any fix regarding this is required.


Dannyps:
#229 by @buster95 does this in a better way, please merge his!

PR #234, opened 2020-03-27
Adding active cases to /all endpoint

Nikhil-Wagh:
It fixes this issue https://github.com/NovelCOVID/API/issues/231


ebwinters:
good idea, I think I'm gonna make this `v2/all`, will merge later today with some changes 👍

PR #243, opened 2020-03-27
add helpful tip for docker users

chriswales95:
Adds a tip for docker users to change their host so the connection isn't refused


ebwinters:
Is this still relevant?

Nekiono:
can be merged

PR #256, opened 2020-03-28
Asynchronous calls for database

chroventer:
- Endb has a simplistic and convenient promise-based API with asynchronous calls. 
- It utilizes its own data serialization methods to assure consistency across multiple storage backends (uses json-buffer#stringify and json-buffer#parse).
- It is also designed to be easily *embeddable* inside modules, which is the perfect functionality for this project. 
- Furthermore, it handles all the JSON types including Buffer, thus requests' data is efficiently regulated and handled. 
- Connection errors are transmitted through, from the adapter to the main instance; consequently, connection errors do not exit or kill the process.

```javascript
// Import/require the package.
const Endb = require('endb');

const endb = new Endb();
const endb = new Endb({
    // One of the following
    uri: 'leveldb://path/to/database',
    uri: 'mongodb://user:pass@localhost:27017/dbname',
    uri: 'mysql://user:pass@localhost:3306/dbname',
    uri: 'postgresql://user:pass@localhost:5432/dbname',
    uri: 'redis://user:pass@localhost:6379',
    uri: 'sqlite://path/to/database.sqlite'
});

// Handles connection errors
endb.on('error', error => console.error('Connection Error: ', error));

await endb.set('foo', 'bar'); // true
await endb.get('foo'); // 'bar'
await endb.has('foo'); // true
await endb.all(); // [ { key: 'foo', value: 'bar' } ]
await endb.delete('foo'); // true
await endb.clear(); // undefined
```

Documentation: [https://endb.js.org](https://endb.js.org)
NPM: [https://npmjs.com/endb](https://npmjs.com/endb)

Note: I have extensively tested the internals, all of the functionality was working perfectly fine.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -1,7 +1,8 @@
 {
   "redis": {
     "host": "localhost",
-    "password": "1234"
+    "password": "1234",
+    "port": 6969
   },
   "keys": {
     "all": "covidapi:all",
--- a/funcs/getAll.js
+++ b/funcs/getAll.js
@@ -29,8 +29,7 @@ var getall = async (keys, redis) => {
   });
   result.updated = Date.now()
   result.active = result.cases - result.deaths - result.recovered
-  const string = JSON.stringify(result);
-  redis.set(keys.all, string);
+  await redis.set(keys.all, result);
   console.log("Updated The Cases", result);
 }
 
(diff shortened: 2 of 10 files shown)


Nekiono:
Dont see the differences apart from another level of abstraction.

Endb rn is at version 0.21.7, which as per semver is not stable.

We're not storing data per se, redis is used here as a cache. Hence, there's not really a point in having support for multiple storage backends.

tl;dr while endb looks like a great aspiring project, I don't think it's suitable for us.

Issue #254, opened 2020-03-28
Mistake on country name


innoving:
Hello,

There is a mistake on this country name : **Coast d'ivoire** .... The right name is **Côte d'ivoire.**
Refence link : https://corona.lmao.ninja/countries
Thanks you advance

buster95:
Working on it

chroventer:
#257

chroventer:
Fix: #258

buster95:
> Fix: #258
Check my comment on your PR

PR #258, opened 2020-03-28
fix: mistake in country's name

chroventer:
Addresses #254.


chroventer:
> You don't add Coast D'Ivoire in possibleNames, web service couldn't found this country

#259 @buster95

EliteDaMyth:
@all-contributors please add @chroventer for bug

PR #266, opened 2020-03-28
feat(v2history): add optional province filter for v2/historical/

cmcleese:
add optional province param for v2 historical.


ebwinters:
check failed check logs

ebwinters:
This really isn't needed as you can simply map the returned JSON from https://corona.lmao.ninja/v2/historical/ and check for `element.country===countryname && element.province===provinceName`

PR #305, opened 2020-03-31
Add counties to v2/jhucsse endpoint

### LINKED ISSUE #289: [Feature Request] API calls for city/county level data
I see that [time series data](https://github.com/CSSEGISandData/COVID-19/tree/master/csse_covid_19_data/csse_covid_19_time_series), which doesn't have city/county level data, is currently being used to back this API. In order to make retrieving more specific data possible, one could use [daily reports](https://github.com/CSSEGISandData/COVID-19/blob/master/csse_covid_19_data/csse_covid_19_daily_reports/03-28-2020.csv) to update a local data source formatted in accordance to current requirements. This would initially require backfilling all the daily reports and ensuring there are no discrepancies between the new data source and the time series data (which shouldn't occur because the the time series data should rely on daily reports). If this feature is added, this API can be used for cases that require specific data. Also, this issue is similar to #56, #285 and #5, which shows people want this feature.

ebwinters:
# What does this PR do?
generally closes #289 
1. Add `v2/jhucsse/counties` endpoint for getting data on every county in the USA
2. Add `v2/jhucsse/counties/county` endpoint for getting data on specific county names.

I have tested the PR locally, but **please** test all `v2/jhucsse` functionality for this PR locally before I merge it, I want some feedback and to make sure the original endpoint is still fine.

# How to test?
1. git remote add pr-305 https://github.com/ebwinters/API
2. git fetch pr-305
3. git checkout -b pr-305-master pr-305/master
4. start a redis server with `redis-server`
5. start node server with `node server.js`
6. go to your local host

### DIFF
--- a/funcs/jhuLocations.js
+++ b/funcs/jhuLocations.js
@@ -38,8 +38,8 @@ const jhudata = async (keys, redis) => {
 	parsed.splice(1).forEach((loc) => {
 		result.push({
 			country: loc[3],
-			province: loc[2] === '' ? null : loc[2],
-			city: loc[1] === '' ? null : loc[1],
+			province: loc[2] || null,
+			city: loc[1] || null,
 			updatedAt: loc[4],
 			stats: {
 				confirmed: loc[7],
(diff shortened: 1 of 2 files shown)


Nekiono:
=== '' ? null : value  
can be shortened to  
value || null  
as empty string is falsy

PR #306, opened 2020-03-31
Query multiple countries for historical data

### LINKED ISSUE #303: [Feature Request] Adding API call for historical data for multiple countries
Having an API call that would pull historical data for multiple countries at once would be extremely useful. Something like https://corona.lmao.ninja/v2/historical/usa,italy,france for names as parameters, or https://corona.lmao.ninja/v2/historical/840,380,250 for ids as parameters, or however you choose to implement it.

AstroCB:
This closes #303 – more specifically, it allows multiple countries to be accessed at the `/v2/historical/query` endpoint, where query can now be a comma-separated list of countries as described in that issue.

It also continues to support provinces in this format: 

`https://corona.lmao.ninja/v2/historical/country1,country2,country3/province1,province2`

where provinces are listed in the same order as the countries (any countries that you don't want provinces for can be listed at the end of the country list, so `country3` in the above list would be returned in full as normal).

I wasn't sure if there was any specific convention to be followed here as I haven't contributed to this project before – let me know if you have any ideas for presenting this data differently.

I'd have updated the docs as well, but they don't seem to be part of this repo.

### DIFF
--- a/server.js
+++ b/server.js
@@ -126,14 +126,24 @@ app.get('/v2/historical/:query/:province?', async (req, res) => {
 	const data = JSON.parse(await redis.get(keys.historical_v2));
 	const { query, province } = req.params;
 	const countries = query.split(',');
-	const provinces = province && province.split(',');
+	const provinces = (province && province.split(',')) || [];
 	let countryData;
-	if (countries.length > 0) {
-		countryData = countries.map((country, i) =>
+	// multiple countries no provinces allowed
+	if (countries.length > 1) {
+		countryData = countries.map((country) =>
 			scraper.historical.getHistoricalCountryDataV2(
 				data,
 				country.toLowerCase(),
-				provinces && provinces[i] && provinces[i].toLowerCase()
+				null
+			) || { err: 'Country not found or doesn\'t have any historical data' }
+		);
+	} else if (provinces.length > 0) {
+		// provinces for one country
+		countryData = provinces.map((prov) =>
+			scraper.historical.getHistoricalCountryDataV2(
+				data,
+				countries[0].toLowerCase(),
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -123,11 +123,24 @@ app.get('/v2/historical/all', async (req, res) => {
 app.get('/v2/historical/:query/:province?', async (req, res) => {
 	const data = JSON.parse(await redis.get(keys.historical_v2));
 	const { query, province } = req.params;
-	const countryData = await scraper.historical.getHistoricalCountryDataV2(
```
AstroCB: The `await` here seemed unnecessary, so I removed it – am I missing something?

ebwinters:
lemme test this tomorrow and then will merge 👍

buster95:
comma sign will be a wrong choice, I think a good choice could be plus sign, because some countries will be using comma in its name
USA+UK

ebwinters:
> comma sign will be a wrong choice, I think a good choice could be plus sign, because some countries will be using comma in its name
> USA+UK

plus is how to link words together by spaces, in an HTML addr it won't even be parsed.

ebwinters:
tbh, I don't really like the idea of making an endpoint that takes comma separated lists of countries and provinces, but the country names that you don't want the provinces for come last. I would prefer you give a list of countries, or a country and list of provinces. That's more intuitive. I'll update the code to do that.

buster95:
> > comma sign will be a wrong choice, I think a good choice could be plus sign, because some countries will be using comma in its name
> > USA+UK
> 
> plus is how to link words together by spaces, in an HTML addr it won't even be parsed.
In URL encode space are encoded like this %20

PR #315, opened 2020-03-31
Remove countryMapping and use only countryUtils file

ebwinters:
this was just unnecessary code


buster95:
tsssssss you finally understood my code 😁, thanks @ebwinters

PR #358, opened 2020-04-02
Add multi-country access to countries endpoint

### LINKED ISSUE #345: Query multiple countries
I was asked about a use case where someone would like to make a clone of http://my-coronavirus-map.netlify.com/ which uses your API for a specific continent which includes multiple countries. Would it be possible to use the `/countries/{query}` endpoint with a delimited list of country codes?

for instance, if I wanted to look up North America, I would be able to do something like `/countries/us,ca,mx` theoretically

In the meantime, I'm just going to do a static list of the countries and filter the data which isn't bad, but thought that might be a nice native API feature.

Thanks

nibble-4bits:
This PR resolves #345 

It adds the feature to be able to query for multiple countries in the `/countries/{query}` endpoint by providing a comma-separated list of country codes _(name, iso2, iso3, id)_ in this format: 

`https://corona.lmao.ninja/countries/country1,country2,country3`

I imitated PR #306, so the order in which the countries get returned is the same order in which the countries are listed in the request param


ebwinters:
nice job on this

colbyfayock:
thank you!

PR #362, opened 2020-04-03
according to ISO 3166, change Belarus iso2 to BY, Chad iso2 to TD

snomile:
reference:
https://www.iso.org/obp/ui/#iso:code:3166:BY
https://www.iso.org/obp/ui/#iso:code:3166:TD


ebwinters:
cool

PR #381, opened 2020-04-03
Add state sorting and /states/state endpoint

benjamin-sommer:


routes/api_worldometers.js
```diff
@@ -71,6 +71,14 @@ router.get('/states', async (req, res) => {
 	res.send(states);
 });
 
+router.get('/states/:query', async (req, res) => {
+	const { query } = req.params;
+	const states = JSON.parse(await redis.get(keys.states));
+	const stateData = states.filter(st => st.state === query);
+	const state = stateData[0] || {};
```
puf17640, line 78: please use
`const state = states.find(st => st.state === query)` to make it shorter

benjamin-sommer: Done

ebwinters:
Don't merge until these things are fixed

Issue #395, opened 2020-04-04
[Feature Request] Adding API call for historical data per Month


fanfaron2004:
Adding API call for historical data  (all and country) per Month

ebwinters:
@buster95 for you

buster95:
ok working on it

PR #413, opened 2020-04-05
[Feature] month query string parameter in historical

buster95:
Add for Documentation and Testing
1. **`/v2/historical?month=3`** _will show all days of march_
1. **`/v2/historical?month=3&lastdays=5`**  _will show the last 5 days of march_

solve #395


ebwinters:
This is just too much extra code for a feature that isn't necessary. I feel like people can filter data on their own if they really need to offer month by month displays.

ebwinters:
Closing for now, will re-open once we have refactored code a bit. Do not like adding so many features to one endpoint in one day.

puf17640:
love the idea but as @ebwinters said, the benefit it would bring is a bit too small to add it to the codebase. i might use your code in my api wrapper though, if i get time to work on it tomorrow.
i would love to discuss this more in the discord though.

PR #418, opened 2020-04-05
add mappings endpoint

payamazadi:
add a new endpoint that pulls in JHU CSSE's recent mapping table that maps their country name into ISO2 and ISO3 and other fields


ebwinters:
this is not necessary dude

payamazadi:
wow, you are fast

i didn't notice the mappings in the utils. sorry.

PR #459, opened 2020-04-07
added world population parsing

puf17640:
also added a fallback to the old method of calculating the population using the cases and CasesPerOneMillion data

data is parsed from: https://livepopulation.com


puf17640:
ahh lets just not do that and keep it simple

PR #463, opened 2020-04-07
Count worldometer table columns

Slethen:
Counts table columns on worldometers, this should prevent downtime when they add new columns. Should be less pressure when more data is added also!


ebwinters:
great

PR #480, opened 2020-04-09
[Code Improvements] Logger for httpErrors

buster95:


puf17640:
@buster95 sorry but shouldn't we use `console.error` for error logging ? 
also shouldnt it be "Cannot load message" instead of "Not message loaded"

PR #489, opened 2020-04-11
master - version changes should not effect old versions structure.

cenkingunlugu:
Closes https://github.com/NovelCOVID/API/issues/488

### DIFF
--- a/routes/api_worldometers.js
+++ b/routes/api_worldometers.js
@@ -38,8 +38,10 @@ router.get('/countries', async (req, res) => {
 router.get('/countries/:query', async (req, res) => {
 	const { query } = req.params;
 	const countries = JSON.parse(await redis.get(keys.countries)).map(fixApostrophe);
+	const isNotStrict = req.query.strict === 'false';
+	const isStrict = !isNotStrict;
 	const countryData = splitQuery(query)
-		.map(country => countryUtils.getCountryWorldometersData(countries, country, true))
+		.map(country => countryUtils.getCountryWorldometersData(countries, country, isStrict))
 		.filter(value => value);
 	if (countryData.length > 0) {
 		res.send(countryData.length === 1 ? countryData[0] : countryData);
--- a/routes/api_worldometers.js
+++ b/routes/api_worldometers.js
@@ -39,7 +39,7 @@ router.get('/countries/:query', async (req, res) => {
 	const { query } = req.params;
 	const countries = JSON.parse(await redis.get(keys.countries)).map(fixApostrophe);
 	const countryData = splitQuery(query)
-		.map(country => countryUtils.getCountryWorldometersData(countries, country, req.query.strict === 'true'))
+		.map(country => countryUtils.getCountryWorldometersData(countries, country, true))
 		.filter(value => value);
 	if (countryData.length > 0) {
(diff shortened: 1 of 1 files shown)


andredippenaar:
It looks like the json structure changes. Instead of being indexed by highest number of cases, it is not alphabetical.

puf17640:
I saw your issue and know why you want to do this, but we can't just remove the strict parameter.

cenkingunlugu:
> I saw your issue and know why you want to do this, but we can't just remove the strict parameter.

I left the not-strict possibility. Please re-check.

puf17640:
hm that might be doable, i will talk to @ebwinters about it.

routes/api_worldometers.js
```diff
@@ -38,8 +38,10 @@ router.get('/countries', async (req, res) => {
 router.get('/countries/:query', async (req, res) => {
 	const { query } = req.params;
 	const countries = JSON.parse(await redis.get(keys.countries)).map(fixApostrophe);
+	const isNotStrict = req.query.strict === 'false';
+	const isStrict = !isNotStrict;
```
puf17640, line 42: please change to
`const isStrict = !(req.query.strict === 'false')`

puf17640:
also if you are implementing this on 1 endpoint, you have to do it for all

routes/api_worldometers.js
cenkingunlugu: done. also for all endpoints using strict queryparam.

puf17640: my bad, change to `const isStrict = req.query.strict !== 'false'`

EliteDaMyth:
i personally dont think its necessary at all. you can easily do this on your end.

cenkingunlugu:
> i personally dont think its necessary at all. you can easily do this on your end.

This is so unprofessional and disappointing. You behave like you developed something for the first time.

When you open source a project, lots of embedded software can include your software in their implementations. This means, you can not change it.

You can add a version, and your old version should not change ever, because it is a breaking change. In your new version you can do whatever you want, but in the older version (which you implemented and given a structure to users) if you are changing it, it means you will break lots of users data.

anyways, i will not try to teach you what you are doing. but your approach is not acceptable.

ebwinters:
@cenkingunlugu what breaking change was made to an older version?

cenkingunlugu:
Did you see my issue? On last 3 weeks when i have a call like countries/netherlands it was returning netherlands results. Till yesterday it returns netherlands antilles.

ebwinters:
@cenkingunlugu yes, we are reopening aother pr to fix this along with adding more tests. I see what happened

puf17640:
> Did you see my issue? On last 3 weeks when i have a call like countries/netherlands it was returning netherlands results. Till yesterday it returns netherlands antilles.

i totally understand you here and as you can see above, i also think this needs to be fixed to not have a breaking change in v1, we didn't purposely change that, it just happened in the means of fixing other stuff. i have already filed a pr now that is fixing it just like you did and will add tests for the edge cases soon with ethan! 👍 

i also totally agree with you saying that the behaviour of elite was unprofessional, your issue is relevant and needs to be addressed and other contributors didnt even get time to view your pr before it was closed down.

PR #493, opened 2020-04-11
strict by default

puf17640:
in memories of #489


ebwinters:
waiting for @EliteDaMyth

PR #495, opened 2020-04-11
[Versioning][Feature Request] V2 States

### LINKED ISSUE #487: States endpoint v2
Move states endpoint to v2 base URL. Also merging this with #481 since yesterday data is available for US states but it doesn't list as a function.

- [ ] Update JSDOC
- [ ] Update Docs
- [ ] Write more tests

ebwinters:
closes #487 

- [x] Update JSDOC
- [x] Update Docs
- [x] Write more tests

### DIFF
--- a/scrapers/getStates.js
+++ b/scrapers/getStates.js
@@ -11,7 +11,6 @@ const parseStateCell = (cell) => {
 		|| '';
 	state = state.trim();
 	if (state.length === 0) {
-		// parse with hyperlink
 		state = cell.children[0].next.children[0].data || '';
 	}
 	return state.trim() || '';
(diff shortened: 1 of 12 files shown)


scrapers/getStates.js
```diff
@@ -11,7 +11,6 @@ const parseStateCell = (cell) => {
 		|| '';
 	state = state.trim();
 	if (state.length === 0) {
-		// parse with hyperlink
 		state = cell.children[0].next.children[0].data || '';
 	}
 	return state.trim() || '';
```
puf17640, line 16: couldnt we use falsy values to cut this down to
`return state.trim() || cell.children[0].next.children[0].data.trim() || '';`

puf17640: also line 42-50 could be reduced down to
`return tableRows.map((row) => {
		const cells = row.children.filter((cell) => cell.name === 'td');
		const stateData = { state: parseStateCell(cells[stateColIndex]) };
		Object.keys(dataColIndexes).forEach((property) => {
			stateData[property] = parseNumberCell(cells[dataColIndexes[property]]);
		});
		return stateData;
	});`

PR #507, opened 2020-04-13
feat(kind): added kubernetes manifests and instructions

lukibahr:
* added instructions for kind (kubernetes in docker) setup
* added manifests for a kubernetes deployment of redis
* added manifests for a kubernetes deployment of the api
* fixed documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,22 @@ NovelCovid/API Documentation can be found [here](https://corona.lmao.ninja/docs/
 5. Replace redis host "localhost" with "redis".
 6. Run command `docker-compose up --build -d`.
 
+## Running in a Kubernetes cluster
+
+The repository contains a very basic setup of the api and redis deployment. HA setups are not considered in the following example.
+
+1. Fork and clone git repository
+2. Install kind for a local cluster using `GO111MODULE="on" go get sigs.k8s.io/kind@v0.7.0` (requires go to be installed)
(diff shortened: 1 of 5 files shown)


EliteDaMyth:
how is it helpful though?

lukibahr:
For running the app in a kubernetes environment. Helpful for serving the app in high-available setups for advanced scaling

puf17640:
we are not using kubernetes to run the API, if we ever need scaling, this would be done using nginx

i still really appreciate the work you have done! have you tried it out also, does it run? how many instances have tried to run?

lukibahr:
tried 2-3 replicas for the api, but havent had the time to run ha-redis setup.

puf17640:
> tried 2-3 replicas for the api, but havent had the time to run ha-redis setup.

The question is, is there a distinct performance increase or other drastic benefits to nginx?

lukibahr:
The performance might be the same, but the major benefit is the autoscaling behaviour and self-healing pattern that kubernetes comes with. How do you scale your apps and services behind your nginx?

puf17640:
@lukibahr you would have to talk to @Nekiono because she is in charge of maintaining the server. she is also from germany. it should be just a default nginx config and the express backend behind it

lukibahr:
Okay, i‘ll talk to her via discord. Let’s keep the branch and mr open till a decision is made

ebwinters:
I don't think we want to use kubernetes, at least not yet. We are looking into a microservices infrastructure which could maybe use it in the future. I will close PR and mark it for later so we can get to it again if needed. Thanks for your work

PR #510, opened 2020-04-14
Initial changes to run unit tests in docker.

sam-perez:
Made some quality of life changes in Dockerfile so that it can reuse installed packages even if you make a code change.
Added .dockerignore to ensure that host machine node_modules do not squash installed node_modules inside the container.
Added `docker-test` command to package.json.
Added `docker-compose.test.yml` to change the command run in the `covid-api-app` container.

Once you've set up your local environment to successfully run the app in docker, you should be able to just run `npm run docker-test`.

It will build the container on every invocation, but docker caching should make the rebuild quick. Note that you can read the exit code of the command, and it will be 0 if all of the tests passed or 1 if there were any failures.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,11 @@ FROM node:13
 RUN mkdir -p /home/container
 WORKDIR /home/container
 
-COPY . /home/container
+# we should only re-run npm install if package.json specifically has changed. use docker build cache otherwise
+COPY ./package.json /home/container/package.json
 RUN npm i
+
+# copy everything else over
+COPY . /home/container
 
 CMD [ "node", "server.js" ]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,0 +1,4 @@
+version: '3.7'
+services:
+  app:
+    command: ["npm", "run", "test"]
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "lint": "eslint '**/*.js' --ignore-pattern node_modules/",
     "lint:fix": "eslint ./**/*.js --fix --ignore-pattern node_modules/",
     "lint:win32": "eslint ./**/*.js --ignore-pattern node_modules/",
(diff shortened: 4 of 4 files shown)


Dockerfile
```diff
@@ -3,7 +3,11 @@ FROM node:13
 RUN mkdir -p /home/container
 WORKDIR /home/container
 
-COPY . /home/container
+# we should only re-run npm install if package.json specifically has changed. use docker build cache otherwise
+COPY ./package.json /home/container/package.json
```
buster95, line 7: This line is unnecessary you can do
COPY ./package.json .

Dockerfile
```diff
@@ -3,7 +3,11 @@ FROM node:13
...
 RUN npm i
 
+# copy everything else over
+COPY . /home/container
```
buster95, line 11: ???????
first you copy only package.json in this line you're copy all files??
I think that you can replace line 7 with
COPY . .   (you can do this because you have a dockerignore file with node_modules then you can copy all)
And delete this line

sam-perez: The reason to do it this way is to differentiate between package.json changes and any other changes with docker build caching.

Check out the documentation here: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

Each command creates a layer, and docker has certain caching strategies to determine if a layer needs to be rerun. The way I've written it, if you make a change to `package.json`, when you build a container, docker will detect that it needs to rerun from line:

`COPY ./package.json /home/container/package.json`

AKA, it will copy `package.json`, then install dependencies, and copy over all of the source code.

But, if you make changes to any other file, docker will detect that it can skip the `COPY ./package.json ...` command, will reuse the previously generated layer, and start from

`COPY . /home/container`

This is important because it will speed up the development flow when you are making changes to test files or other source files, but not touching any dependencies. It speeds up test run by at least a few seconds each time you run it.

Does this make sense?

puf17640: I will look over this soon!

EliteDaMyth:
thanks for putting in the time and effort. we have talked alot about this and came to a decision that its unnecessary. sorry for wasting your time!

sam-perez:
No problem, what was the conversation? Also, are you guys still trying to split out the scraper process from the server?

PR #517, opened 2020-04-15
[Proposal] Dockerize scraper

sam-perez:
Picking up where I left off with dockerizing unit tests.

### GOALS:

1) Split up the scraper job from the web server.
2) Enable docker development flow.

### DESIGN:

We were able to reuse most of the existing code. The main changes were:

- Refactoring the scraper job out of `server.js` and creating a `scraper.js` entrypoint for the scraper.
- Creating a second container for the scraper that uses the `scraper.js` entrypoint. It is able to reuse the `Dockerfile` for the web server container, and also uses a lot of the same configuration in the `docker-compose` files.
- Created a command in `package.json` to both start the production containers and also run the unit tests. Look at `docker-start` and `docker-test`.

### Notes:

I've moved all of the base docker-compose components that are shared between prod and test into `docker-compose.yml`. Using the yml merging feature of `docker-compose`, there are two new docker-compose yml configs for test and prod. The configs use different names for each container in test and prod, as well as the prod config defining the scraper container. The redis containers in test and prod use different volumes so that unit tests do not interact with the main redis store, and vice versa.

The scraper container is not run at all during unit tests. Instead, we've added a `mochaGlobalSetup.js` file, which contains all of the setup we need to run before any mocha tests are run. This is where we run the initial scrape job, but not before flushing the redis instance to ensure that a previous test run is cleaned up before running the current test run.

Note that the docker file seemingly has a redundant copy of `package.json` before copying over all of the contents of the directory. This is intentional, and we are leveraging the caching of docker layers when docker is building images in order to decrease the time between tests runs when code is changed but we have not made any changes to any node dependencies and don't need to re-run `npm i`. Look [here](https://github.com/NovelCOVID/API/pull/510#discussion_r407885729) for more details.

### TODO

We would need to update the readme to reflect the new way of running the app in docker.

Do we want to flush redis in all environments (aka, if someone is developing locally and not in docker)? Potentially, we could set an environment variable that controls this behavior and turn it on in docker, while leaving it off by default otherwise.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -1,0 +1,25 @@
+version: '3.7'
+services:
+  redis:
+    container_name: covid-api-redis
+    volumes:
+      - redis-data:/data
+
+  scraper:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: covid-api-app-scraper
+    networks:
+      - redis-net
+    depends_on:
+      - redis
+    command: ["node", "scrape.js"]
+
+  app:
+    ports:
+      - 30001:3000
+    container_name: covid-api-app
+
+volumes:
+  redis-data:
--- a/routes/instances.js
+++ b/routes/instances.js
@@ -30,6 +30,15 @@ module.exports = {
 		getWorldometerPage,
 		getStates,
 		jhuLocations,
-		historical
+		historical,
+		executeScrapeJob: async () => {
+			await Promise.all([
+				getWorldometerPage(keys, redis),
+				getStates(keys, redis),
+				jhuLocations.jhudataV2(keys, redis),
+				historical.historicalV2(keys, redis),
+				historical.getHistoricalUSADataV2(keys, redis)
+			]);
+		}
 	}
 };
--- a/scrape.js
+++ b/scrape.js
@@ -1,0 +1,10 @@
+const { config, scraper: { executeScrapeJob } } = require('./routes/instances');
+
+const logWithDate = (message) => console.log(`[${new Date().toISOString()}]: ${message}`);
+
+logWithDate('Starting initial scrape job...');
+executeScrapeJob().then(() => logWithDate('Initial scrape job finished.'));
+setInterval(() => {
+	logWithDate('Scrape job waking up and executing...');
+	executeScrapeJob().then(() => logWithDate('Finished scrape job.'));
+}, config.interval);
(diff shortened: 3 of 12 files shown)


tests/mochaGlobalSetup.js
```diff
@@ -0,0 +1,10 @@
+const { scraper: { executeScrapeJob }, redis } = require('../routes/instances');
+
+// eslint-disable-next-line
+before(async () => {
+	await redis.flushall();
```
puf17640, line 5: really not a good idea to flush redis before trying to scrape all the data, what will you do when you flush and the whole scraping fails? nginx will save your ass because it will serve the cache, but you will have to wait at least 10 minutes until it is updated again, IF the scraping doesnt fail again

scrape.js
```diff
@@ -0,0 +1,10 @@
+const { config, scraper: { executeScrapeJob } } = require('./routes/instances');
+
+const logWithDate = (message) => console.log(`[${new Date().toISOString()}]: ${message}`);
```
puf17640, line 3: a logger exists now for that purpose

routes/instances.js
```diff
@@ -30,6 +30,15 @@ module.exports = {
 		getWorldometerPage,
```
puf17640, line 30: why return all of them if you now only need 1 function (executeScrape)

tests/mochaGlobalSetup.js
sam-perez: This is for testing specifically. If scraping fails, tests fail (dy design?)

routes/instances.js
sam-perez: Some of the functions are referenced in other places.

puf17640: good argument, not gonna lie

tests/mochaGlobalSetup.js
puf17640: okay okay, i didnt look at the filename, makes sense now haha

docker-compose.prod.yml
```diff
@@ -0,0 +1,25 @@
...
+    container_name: covid-api-app-scraper
+    networks:
+      - redis-net
+    depends_on:
+      - redis
+    command: ["node", "scrape.js"]
+
+  app:
+    ports:
+      - 30001:3000
```
ebwinters, line 21: don't you mean 3001

buster95:
@ebwinters I'm working on this PR

puf17640:
since we are in the state of dockerization, is this closeable? @ebwinters @buster95

ebwinters:
Ye

PR #518, opened 2020-04-16
NYT Integration

vincentvb:
This PR adds the following routes:

`/v2/nyt/states`
`/v2/nyt/counties`
`/v2/nyt/usa`

This allows users to optionally query for NYT data rather than Johns Hopkins or Worldometer if they so choose.

Some extra thoughts:
- The state and county routes currently support looking up ALL state/county data, or data for a single state/county. Maybe we want to include the option to look up multiple states/counties at once? 
- Do we want to add functionality to the `node-api` package as well?
- We'll need to update the deployed version of the `config.json` to add the `nyt_interval` before any changes are deployed

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -5,5 +5,6 @@
     "port": "6379"
   },
   "port": 3000,
-  "interval": 600000
+  "interval": 600000,
+  "nyt_interval": 72e6
 }
--- a/config.example.json
+++ b/config.example.json
@@ -6,5 +6,5 @@
   },
   "port": 3000,
   "interval": 600000,
-  "nyt_interval": 72e6
+  "nyt_interval": 36e5
 }
--- a/config.keys.json
+++ b/config.keys.json
@@ -8,7 +8,7 @@
   "jhu_v2": "covidapi:jhu_v2",
   "historical_v2": "covidapi:historical_v2",
   "historical_v2_USA": "covidapi:historical_v2_USA",
-  "nyt_counties": "covidapi:nyt:counties",
-  "nyt_states": "covidapi:nyt:states",
-  "nyt_USA": "covidapi:nyt:nation"
+  "nyt_counties": "covidapi:nyt_counties",
+  "nyt_states": "covidapi:nyt_states",
+  "nyt_USA": "covidapi:nyt_usa"
 }
--- a/config.keys.json
+++ b/config.keys.json
@@ -7,5 +7,8 @@
   "yesterday_states": "covidapi:yesterday_states",
   "jhu_v2": "covidapi:jhu_v2",
   "historical_v2": "covidapi:historical_v2",
-  "historical_v2_USA": "covidapi:historical_v2_USA"
+  "historical_v2_USA": "covidapi:historical_v2_USA",
+  "nyt_counties": "covidapi:nyt:counties",
+  "nyt_states": "covidapi:nyt:states",
+  "nyt_nationwide": "covidapi:nyt:nation"
 }
--- a/config.keys.json
+++ b/config.keys.json
@@ -10,5 +10,5 @@
   "historical_v2_USA": "covidapi:historical_v2_USA",
   "nyt_counties": "covidapi:nyt:counties",
   "nyt_states": "covidapi:nyt:states",
-  "nyt_nationwide": "covidapi:nyt:nation"
+  "nyt_USA": "covidapi:nyt:nation"
 }
--- a/routes/api_nyt.js
+++ b/routes/api_nyt.js
@@ -2,7 +2,7 @@
 const router = require('express').Router();
 const { nytCounties, nytStates, nytNationwide } = require('../utils/nyt_cache');
 
-router.get('/nyt/states', async (req, res) => {
+router.get('/v2/nyt/states', async (req, res) => {
 	const data = nytStates();
 	const { state: queryState } = req.query;
 	if (queryState) {
@@ -16,7 +16,7 @@ router.get('/nyt/states', async (req, res) => {
 	}
 });
 
-router.get('/nyt/counties', async (req, res) => {
+router.get('/v2/nyt/counties', async (req, res) => {
 	const data = nytCounties();
 	const { county: queryCounty } = req.query;
 	if (queryCounty) {
@@ -30,7 +30,7 @@ router.get('/nyt/counties', async (req, res) => {
 	}
 });
 
-router.get('nyt/nation_wide', async (req, res) => {
+router.get('/v2/nyt/nation_wide', async (req, res) => {
 	res.send(nytNationwide());
 });
 
--- a/routes/api_nyt.js
+++ b/routes/api_nyt.js
@@ -6,7 +6,8 @@ router.get('/v2/nyt/states', async (req, res) => {
 	const data = nytStates();
 	const { state: queryState } = req.query;
 	if (queryState) {
-		const stateData = data.filter(({ state }) => queryState === state);
+		const stateArr = queryState.split(/,[\s+]?/);
+		const stateData = data.filter(({ state }) => stateArr.includes(state));
 		// eslint-disable-next-line no-unused-expressions
 		stateData.length > 0
 			? res.send(stateData)
@@ -20,7 +21,8 @@ router.get('/v2/nyt/counties', async (req, res) => {
 	const data = nytCounties();
 	const { county: queryCounty } = req.query;
 	if (queryCounty) {
-		const countyData = data.filter(({ county }) => queryCounty === county);
+		const countyArr = queryCounty.split(/,[\s+?]/);
+		const countyData = data.filter(({ county }) => countyArr.includes(county));
 		// eslint-disable-next-line no-unused-expressions
 		countyData.length > 0
 			? res.send(countyData)
@@ -30,7 +32,7 @@ router.get('/v2/nyt/counties', async (req, res) => {
 	}
 });
 
-router.get('/v2/nyt/nation_wide', async (req, res) => {
+router.get('/v2/nyt/usa', async (req, res) => {
 	res.send(nytNationwide());
 });
 
--- a/routes/api_nyt.js
+++ b/routes/api_nyt.js
@@ -2,8 +2,8 @@
 const router = require('express').Router();
 const { nytCounties, nytStates, nytNationwide } = require('../utils/nyt_cache');
 
-router.get('/v2/nyt/states', async (req, res) => {
-	const { state: queryState } = req.query;
+router.get('/v2/nyt/states/:state', async (req, res) => {
+	const { state: queryState } = req.params;
 	const data = nytStates();
 	if (queryState) {
 		const stateArr = queryState.split(/,[\s+]?/).map((state) => state.toLowerCase());
@@ -17,8 +17,8 @@ router.get('/v2/nyt/states', async (req, res) => {
 	}
 });
 
-router.get('/v2/nyt/counties', async (req, res) => {
-	const { county: queryCounty } = req.query;
+router.get('/v2/nyt/counties/:county', async (req, res) => {
+	const { county: queryCounty } = req.params;
 	const data = nytCounties();
 	if (queryCounty) {
 		const countyArr = queryCounty.split(/,[\s+?]/).map((county) => county.toLowerCase());
--- a/routes/api_nyt.js
+++ b/routes/api_nyt.js
@@ -3,8 +3,8 @@ const router = require('express').Router();
 const { nytCounties, nytStates, nytNationwide } = require('../utils/nyt_cache');
 
 router.get('/v2/nyt/states', async (req, res) => {
-	const data = nytStates();
 	const { state: queryState } = req.query;
+	const data = nytStates();
 	if (queryState) {
 		const stateArr = queryState.split(/,[\s+]?/);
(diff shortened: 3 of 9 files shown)


config.keys.json
```diff
@@ -7,5 +7,8 @@
   "yesterday_states": "covidapi:yesterday_states",
   "jhu_v2": "covidapi:jhu_v2",
   "historical_v2": "covidapi:historical_v2",
-  "historical_v2_USA": "covidapi:historical_v2_USA"
-}
+  "historical_v2_USA": "covidapi:historical_v2_USA",
+  "nyt_counties": "covidapi:nyt:counties",
+  "nyt_states": "covidapi:nyt:states",
+  "nyt_nationwide": "covidapi:nyt:nation"
```
ebwinters, line 13: might want to call this nyt_USA just to keep it how it has been

server.js
```diff
@@ -17,8 +17,13 @@ const execAll = async () => {
 	app.emit('scrapper_finished');
 };
 
+const execNyt = () => scraper.nytData(keys, redis);
+
 execAll();
+execNyt();
+
 setInterval(execAll, config.interval);
+setInterval(execNyt, config.nyt_interval);
```
ebwinters, line 26: would need Ravy to update config.json - talk to her on this

routes/api_nyt.js
```diff
@@ -0,0 +1,37 @@
...
+		// eslint-disable-next-line no-unused-expressions
+		countyData.length > 0
+			? res.send(countyData)
+			: res.status(404).send({ message: 'County not found or no data found for county' });
+	} else {
+		res.send(data);
+	}
+});
+
+router.get('/v2/nyt/nation_wide', async (req, res) => {
```
ebwinters, line 33: consider changing to /america or /usa instead of nation_wide

config.example.json
```diff
@@ -5,5 +5,6 @@
     "port": "6379"
   },
   "port": 3000,
-  "interval": 600000
-}
+  "interval": 600000,
+  "nyt_interval": 72e6
```
EliteDaMyth, line 9: 72e6 milliseconds is like 1200 minutes i.e updates every 1200 minutes (20 hours.) Are you sure this is the value you want?

ebwinters: They only update their data once a day

vincentvb: After talking on Discord, changed the interval to 1 hour.

config.keys.json
```diff
@@ -7,5 +7,8 @@
   "yesterday_states": "covidapi:yesterday_states",
   "jhu_v2": "covidapi:jhu_v2",
   "historical_v2": "covidapi:historical_v2",
-  "historical_v2_USA": "covidapi:historical_v2_USA"
-}
+  "historical_v2_USA": "covidapi:historical_v2_USA",
+  "nyt_counties": "covidapi:nyt:counties",
```
puf17640, line 11: Can we stay at the _ format ? covidapi:nyt_countries ?

routes/api_nyt.js
```diff
@@ -0,0 +1,39 @@
+// eslint-disable-next-line new-cap
+const router = require('express').Router();
+const { nytCounties, nytStates, nytNationwide } = require('../utils/nyt_cache');
+
+router.get('/v2/nyt/states', async (req, res) => {
+	const data = nytStates();
+	const { state: queryState } = req.query;
```
puf17640, line 7: to fit the other endpoints, move object deconstruction for query parameters to the first line, then initialize data

routes/api_nyt.js
```diff
@@ -0,0 +1,39 @@
...
+			? res.send(stateData)
+			: res.status(404).send({ message: 'State not found or no data found for state' });
+	} else {
+		res.send(data);
+	}
+});
+
+router.get('/v2/nyt/counties', async (req, res) => {
+	const data = nytCounties();
+	const { county: queryCounty } = req.query;
```
puf17640, line 22: same as above

routes/api_nyt.js
```diff
@@ -0,0 +1,39 @@
...
+		countyData.length > 0
+			? res.send(countyData)
+			: res.status(404).send({ message: 'County not found or no data found for county' });
+	} else {
+		res.send(data);
+	}
+});
+
+router.get('/v2/nyt/usa', async (req, res) => {
+	res.send(nytNationwide());
```
puf17640, line 36: 1 liner

scrapers/nytData.js
```diff
@@ -0,0 +1,36 @@
+const axios = require('axios');
+const csv = require('csvtojson');
+const logger = require('../utils/logger');
+const { updateCache } = require('../utils/nyt_cache');
+
+const US_COUNTY_DATA_URL = 'https://raw.githubusercontent.com/nytimes/covid-19-data/master/us-counties.csv';
+const US_STATE_DATA_URL = 'https://raw.githubusercontent.com/nytimes/covid-19-data/master/us-states.csv';
+const US_NATION_WIDE_DATA_URL = 'https://raw.githubusercontent.com/nytimes/covid-19-data/master/us.csv';
+const REDIS_KEYS = ['nyt_counties', 'nyt_states', 'nyt_USA'];
```
puf17640, line 9: you are using different redis keys than you propose in the config

scrapers/nytData.js
```diff
@@ -0,0 +1,36 @@
+const axios = require('axios');
+const csv = require('csvtojson');
+const logger = require('../utils/logger');
+const { updateCache } = require('../utils/nyt_cache');
+
+const US_COUNTY_DATA_URL = 'https://raw.githubusercontent.com/nytimes/covid-19-data/master/us-counties.csv';
+const US_STATE_DATA_URL = 'https://raw.githubusercontent.com/nytimes/covid-19-data/master/us-states.csv';
+const US_NATION_WIDE_DATA_URL = 'https://raw.githubusercontent.com/nytimes/covid-19-data/master/us.csv';
```
puf17640, line 8: cant the data ulrs be represented as a base and then having their ending appended respectively?

`base = "https://raw.githubusercontent.com/nytimes/covid19-data/master/"`

server.js
puf17640: we would need to anway since my last pr changed the redis keys, or not?

scrapers/nytData.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Retrieves NYT data from github and stores it in redis
+ * @param {Object} keys Redis keys sourced from configurations
+ * @param {Object} redis Redis instance
+ */
+const nytData = async (keys, redis) => {
+	try {
+		const _resolveData = async (url, index) => {
+			const { data } = await axios.get(url);
```
puf17640, line 19: maybe add checking response code like we do in the other scraper code, i dont know, i like your way too, but we kinda should agree on one way

scrapers/nytData.js
vincentvb: I think the label I attached is confusing here because I'm referencing the key in the `config.keys.json` file that retrieves the Redis key itself 😆 . Should probably label it `REDIS_CONFIG_KEYS` or something like that, but open to suggestions.

scrapers/nytData.js
vincentvb: Not sure what you mean exactly, can you link to the code in the scraper code that you're referring to?

scrapers/nytData.js
puf17640: no you were literally using "nyt_counties" in this file and using "nyt:counties" in config.keys.json. but i see you already fixed it! nice! i would still propose to not have them hardcoded in your scraper code. 
please have a look at the other scraper code, the server.js file passes the redis keys when calling the scrape function

scrapers/nytData.js
puf17640: okay seems i was the one to ditch this in my last refactoring, so forget about it, i love how did that object destructing

vincentvb:
@puf17640 Yeah I decided to use query params rather than the url path to extract specific states/counties. Best practice is usually to go with a query param if you're filtering out a large data set, which is what's happening here. The path is usually used if you have an id for a database lookup for instance.

scrapers/nytData.js
puf17640: seems i was the one to ditch this in the last refactor haha, forget about it, maybe make a try catch around (like in getWorldometerPage and then call the logger if an err occurs)
love the object deconstruction!!

scrapers/nytData.js
puf17640: ahhhh i see!

puf17640:
> @puf17640 Yeah I decided to use query params rather than the url path to extract specific states/counties. Best practice is usually to go with a query param if you're filtering out a large data set, which is what's happening here. The path is usually used if you have an id for a database lookup for instance.

I only partly agree here. I get why you implemented it that way but I just don't really see why having 2 seperate working endpoint types would be in any way beneficial, since we have been going with the current approach for quite some time now. I also can't think of any benefit this would bring, making it best practice. It makes query strings longer by appending an unneccesary key (?county=), which makes them harder to read and debug. 

still open for discussion if you know stuff i dont 👍

vincentvb:
> > @puf17640 Yeah I decided to use query params rather than the url path to extract specific states/counties. Best practice is usually to go with a query param if you're filtering out a large data set, which is what's happening here. The path is usually used if you have an id for a database lookup for instance.
> 
> I only partly agree here. I get why you implemented it that way but I just don't really see why having 2 seperate working endpoint types would be in any way beneficial, since we have been going with the current approach for quite some time now. I also can't think of any benefit this would bring, making it best practice. It makes query strings longer by appending an unneccesary key (?county=), which makes them harder to read and debug.
> 
> still open for discussion if you know stuff i dont 👍

As with most REST standards, best practices are just general guidelines that developers are encouraged to use. There's no technical reason why you can't use the query path, it's just generally frowned upon for this use case.

Here's a stack overflow post that you might be interested in: https://stackoverflow.com/questions/30967822/when-do-i-use-path-params-vs-query-params-in-a-restful-api

I think the strongest case against using query params is that it's not being used by the rest of the api, so I'm open to refactoring if you all want to maintain this standard.

puf17640:
> > > @puf17640 Yeah I decided to use query params rather than the url path to extract specific states/counties. Best practice is usually to go with a query param if you're filtering out a large data set, which is what's happening here. The path is usually used if you have an id for a database lookup for instance.
> > 
> > 
> > I only partly agree here. I get why you implemented it that way but I just don't really see why having 2 seperate working endpoint types would be in any way beneficial, since we have been going with the current approach for quite some time now. I also can't think of any benefit this would bring, making it best practice. It makes query strings longer by appending an unneccesary key (?county=), which makes them harder to read and debug.
> > still open for discussion if you know stuff i dont 👍
> 
> As with most REST standards, best practices are just general guidelines that developers are encouraged to use. There's no technical reason why you can't use the query path, it's just generally frowned upon for this use case.
> 
> Here's a stack overflow post that you might be interested in: https://stackoverflow.com/questions/30967822/when-do-i-use-path-params-vs-query-params-in-a-restful-api
> 
> I think the strongest case against using query params is that it's not being used by the rest of the api, so I'm open to refactoring if you all want to maintain this standard.

really good stackoverflow post! but after reading it, im even more sure we should keep it like we do, since the countrys name is unique and acts as our "ID" even if we dont have indices and db querying.

vincentvb:
> > > > @puf17640 Yeah I decided to use query params rather than the url path to extract specific states/counties. Best practice is usually to go with a query param if you're filtering out a large data set, which is what's happening here. The path is usually used if you have an id for a database lookup for instance.
> > > 
> > > 
> > > I only partly agree here. I get why you implemented it that way but I just don't really see why having 2 seperate working endpoint types would be in any way beneficial, since we have been going with the current approach for quite some time now. I also can't think of any benefit this would bring, making it best practice. It makes query strings longer by appending an unneccesary key (?county=), which makes them harder to read and debug.
> > > still open for discussion if you know stuff i dont 👍
> > 
> > 
> > As with most REST standards, best practices are just general guidelines that developers are encouraged to use. There's no technical reason why you can't use the query path, it's just generally frowned upon for this use case.
> > Here's a stack overflow post that you might be interested in: https://stackoverflow.com/questions/30967822/when-do-i-use-path-params-vs-query-params-in-a-restful-api
> > I think the strongest case against using query params is that it's not being used by the rest of the api, so I'm open to refactoring if you all want to maintain this standard.
> 
> really good stackoverflow post! but after reading it, im even more sure we should keep it like we do, since the countrys name is unique and acts as our "ID" even if we dont have indices and db querying. it makes it more predictable

The distinction though is we're not actually retrieving the data based on the "id", we're filtering the data that we're retrieving based on the "id". No matter what state, county, country, etc. that we provide we're retrieving the same data, aka "resource", from Redis. That's the difference between retrieving a different resource, which is what the extra query path is supposed to achieve, vs. filtering a resource, which is what the query param is supposed to achieve.

puf17640:
yeah I got what you meant, it's just that we have been going with this for a while and its all just opinions but i think it makes sense to leave it with query variables since the country name, as said before, is a unique value and no matter how exactly we retrieve the data for that request, it is more or less treated like an id

since both approaches are doable and should work fine, i would want @ebwinters and @EliteDaMyth to make this decision

ebwinters:
Possible improvements moving forward
- [ ] query params instead of path params (would be entire API)
- [ ] allowing for no white space between commas
- [ ] structuring data more similarly to other endpoints (i.e. one entry per state with array of dates)

PR #524, opened 2020-04-16
added an endpoint to query cloudflare statistics 

puf17640:
this does not pass linting, should pass all tests tho, @ebwinters  please pull fast and merge your shit so this isnt in the code for too long

script.js could look like this
```
// Read the CSRF token from the <meta> tag
var token = document.querySelector('meta[name="csrf-token"]').getAttribute('content')

// Make a request using the Fetch API
fetch('/private/cloudflare', {
  credentials: 'same-origin', // <-- includes cookies in the request
  headers: {
    'CSRF-Token': token // <-- is the csrf token as a header
  },
  method: 'POST'
}) 
```

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -5,5 +5,6 @@
     "port": "6379"
   },
   "port": 3000,
-  "interval": 600000
+  "interval": 600000,
+  "cfApiKey": "YlJFwDtF7UzyVMFZLXGf6VnCDVAxjAgImwwlEATH"
 }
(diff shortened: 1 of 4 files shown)


puf17640:
fuck looks like i forgot to -s when installing node-fetch

puf17640:
#### added comments in code but here:

you have to change index.html to index.ejs because we will use ejs as view engine
add this to head of index.ejs so you can use it in the script.js as i said above:
<meta name="csrf-token" content="<%= csrfToken %>">

PR #527, opened 2020-04-16
worked on the deprecation thing, we now send more helpful responses

puf17640:


Gabplay:
Cant you redirect requests to /v2? Otherwise it will break everyone who was using it...

puf17640:
> Cant you redirect requests to /v2? Otherwise it will break everyone who was using it...

This is not how deprecating old endpoints works. 
The data structure is not consistent between v1 and v2 endpoints, thus the change in major version. 
We added Warning header (according to RFC) and the link to our docs in every response.

puf17640:
thank god

PR #537, opened 2020-04-17
swagger continent

### LINKED ISSUE #516: Swagger documentation v2/continents
@puf17640 pls do

puf17640:
closes #516


EliteDaMyth:
why were the integer changed to numbers?

puf17640:
> why were the integer changed to numbers?

first of all it was inconsistent before and second of all there is no integer, neither in js nor in json

PR #560, opened 2020-04-20
country list in continents endpoint

### LINKED ISSUE #539: add countries list to /continents endpoint too (now only on specified continent)
to have consistent modeling of continent object, could be beneficial to applications using the api

puf17640:
closes #539


EliteDaMyth:
yo @puf17640 maybe add tests for it as well?

PR #593, opened 2020-04-25
Github action timezone USA/NY

ebwinters:
Fixes JHU unit test issue as long as they update data by when they say they will


buster95:
@ebwinters I solved all this in my PR

PR #596, opened 2020-04-25
Docker Integration 😎

buster95:


docker-compose.prod.yml
```diff
@@ -0,0 +1,42 @@
...
+services:
+  redis:
+    container_name: covid-api-redis
+    command:
+      [
+        "redis-server",
+        "--appendonly",
+        "yes",
+        "--requirepass",
+        "123456",
```
puf17640, line 12: @buster95 is it possible to use the value specified in .env here?

PR #598, opened 2020-04-26
Some fixed

buster95:


config/example.env
```diff
@@ -1,4 +0,0 @@
-REDIS_HOST=redis
-REDIS_PORT=6379
-REDIS_PASSWORD=123456
-SERVER_PORT=3000
```
ebwinters, line 4: Why remove the example config? We may need it for dev setup instructions in the future.

buster95: Because we have two example files one in the project root and this

buster95: Because we have two example files one in the project root and this

buster95: Because we have two example files one in the project root and this

buster95: Because we have two example files one in the project root and this

buster95: Because we have two example files one in the project root and this

buster95: Because we have two example files one in the project root and this

buster95: Because we have two example files one in the project root and this

buster95: Because we have two example files one in the project root and this

PR #601, opened 2020-04-26
github actions workflow and prod using published image

ebwinters:


puf17640:
typo fixed, thank me later, that was my work in terms of docker dev

PR #617, opened 2020-04-30
Stat Section in Home Page

buster95:
## In This PR
1. Stat Section in Home Page
image
2. HTML tags not closed properly

### DIFF
--- a/public/index.ejs
+++ b/public/index.ejs
@@ -1,12 +1,13 @@
 <!DOCTYPE html>
 <html>
+
 <head>
 	<meta charset="utf-8">
 	<meta name="theme-color" content="#303136" />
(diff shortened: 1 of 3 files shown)


public/index.ejs
```diff
@@ -57,19 +59,31 @@
...
+		<h3 class="title">Covid19 Global Stats</h3>
+
+		<div class="grid-container">
+			<div class="grid-item">
+				<div id="chartInfectionCurve"></div>
+			</div>
+
+			<!-- <div class="grid-item" style="background: blue;">
+				corrales
+			</div> -->
```
ebwinters, line 84: remove dis

puf17640, line 84: # corrales

buster95: sorry I was testing

PR #650, opened 2020-05-08
save empty Worldometer values as null instead of zero

andrewvanhemelrijck:
Worldometer data resets for the day at GMT midnight and all "today" values reset to empty. The api returns 0 when cases actually haven't been reported yet. This change returns null instead of zero in those cases. What do you think?

### DIFF
--- a/scrapers/getWorldometers.js
+++ b/scrapers/getWorldometers.js
@@ -47,7 +47,7 @@ const mapRows = (_, row) => {
 				break;
 
 			default:
-				country[columns[index - 1]] = cell.text() === '' ? null : (parseInt(cell.text().replace(/(\n|,)/g, '')) || 0);
+				country[columns[index - 1]] = isNaN(parseInt(cell.text().replace(/(\n|,)/g, ''))) ? null : parseInt(cell.text().replace(/(\n|,)/g, ''));
 		}
 	});
 	return country;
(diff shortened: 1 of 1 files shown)


scrapers/getWorldometers.js
```diff
@@ -47,7 +47,7 @@ const mapRows = (_, row) => {
 				break;
 
 			default:
-				country[columns[index - 1]] = parseInt(cell.text().replace(/(\n|,)/g, '')) || 0;
+				country[columns[index - 1]] = cell.text() === '' ? null : (parseInt(cell.text().replace(/(\n|,)/g, '')) || 0);
```
puf17640, line 50: you could've basically just removed `|| 0` in the old line 50

andrewvanhemelrijck: parseInt gives me NaNs when I do that

andrewvanhemelrijck: Just noticed that although most missing values are empty strings, some are actually a space. Maybe this would be better?
```
isNaN(parseInt(cell.text().replace(/(\n|,)/g, ''))) ? null : parseInt(cell.text().replace(/(\n|,)/g, ''))
```

ebwinters:
Something you gotta think about here is expected values. Doing this could break a lot of people’s code without knowing. I think it is something we put into v3 and don’t just add in

andrewvanhemelrijck:
That's a good point.

puf17640:
we will be sure to address this when planning out the next version of the API

andrewvanhemelrijck:
no problem, happy to help!

PR #651, opened 2020-05-08
Government specific data

ebwinters:
`/v2/gov` endpoint:
- Returns a list of countries whose government data we support

`/v2/gov/Canada` endpoint:
- Data on Canadian provinces scraped from gov site


puf17640:
lgtm, if it works 😆 

i will check it out later today

PR #662, opened 2020-05-12
Austria Gov + refactoring

puf17640:
**WORK IN PROGRESS**


puf17640:
**TO BE ADDED**
- deaths per province
- recovered per province
- deaths by sex
- deaths by age

PR #682, opened 2020-05-16
Add vietnam gov data 

phamleduy04:

PR #683, opened 2020-05-16
Phamleduy04 master

puf17640:
#682 

image


scrapers/govScrapers/getVietnam.js
```diff
@@ -0,0 +1,38 @@
...
+	});
+	return city;
+};
+
+/**
+ * Scrapes Vietnam government site and fills array of data from table
+ */
+const vietnamData = async () => {
+	try {
+		const html = cheerio.load((await axios.default({ method: 'GET', url: 'http://ncov.moh.gov.vn/', httpsAgent: new (require('https').Agent)({ rejectUnauthorized: false }) })).data);
```
puf17640, line 30: For anyone interested, we are setting rejectUnauthorized to false here so we can request the vietnamese government site without getting SSL errors because their SSL chain is invalid. Most browsers fill that chain themselves but axios does not do that of course.

phamleduy04:
Nice i love it

package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "NovelCOVID-API",
+  "name": "novelcovid-api",
```
ebwinters, line 2: Why are we changing this?

tests/api_gov.spec.js
```diff
@@ -366,3 +366,25 @@ describe('TESTING /v2/gov/india', () => {
             });
     });
 });
+
+describe('TESTING /v2/gov/vietnam', () => {
```
ebwinters, line 370: Consider adding another test to test that the data returned doesn't just have properties set, but has then set correctly. For example, making sure cases >= 0.

package-lock.json
puf17640: this was changed long ago, but the package-lock.json hasnt been updated since then. its to be conform with package.json naming rules.

tests/api_gov.spec.js
puf17640: we should do that for all gov tests yeah, will do

PR #691, opened 2020-05-17
allowNull for gov and states

### LINKED ISSUE #690: [Suggestion] Support for null to 0 in government data
I don't know about other government end points, but at least on `/v2/gov/India` there are some null values at the end of the response, it'll save me (and others too) from manually having to check if the value is null or not, I think the default value, i.e. 0 is appropriate enough than to have our app crashing. 
Thanks in advance!

puf17640:
closes #690


puf17640:
aight

PR #696, opened 2020-05-18
Daily Cases for Global historical data, GET /v2/historical/all

buster95:
Working...

### DIFF
--- a/scrapers/historical.js
+++ b/scrapers/historical.js
@@ -172,12 +172,13 @@ const getHistoricalAllDataV2 = (data, lastdays = 30) => {
 	const recovered = {};
 	const dailyCases = {};
 	data.forEach(country => {
-		let lastkey = undefined;
+		let lastkey;
 		Object.keys(country.timeline.cases).slice(lastdays * -1).forEach(key => {
 			/* eslint no-unused-expressions: ["error", { "allowTernary": true }] */
 			cases[key] ? cases[key] += country.timeline.cases[key] : cases[key] = country.timeline.cases[key];
 			deaths[key] ? deaths[key] += country.timeline.deaths[key] : deaths[key] = country.timeline.deaths[key];
 			recovered[key] ? recovered[key] += country.timeline.recovered[key] : recovered[key] = country.timeline.recovered[key];
+			// eslint-disable-next-line no-sequences, no-unused-expressions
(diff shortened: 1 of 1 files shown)


puf17640:
what about dailyDeaths and dailyRecovered ? also, should we really do this? @ebwinters

ebwinters:
yeah I don't really know about this, as it isn't data directly provided by he source and there's no real need

buster95:
I was making this because, I'd like to add a graph to see the infection curve, something like this
image

buster95:
> what about dailyDeaths and dailyRecovered ? also, should we really do this? @ebwinters

eventually I could add dailyDeaths and dailyRecovered

puf17640:
yeah but the point is we shouldnt really add to much to it that isnt provided in the first place to keep it more an official thing, for the countries endpoints I did it because worldometers was already providing a few of the "perOneMillion" values and so it was pretty much implied to have those as well for the other values (even more after we got population value per country)

buster95:
data is provided only subtracts the day from the day before, if you check other data is the same thing

GliderCodes:
something I'm currently using to get the daily data
```typescript
function getDaily(data) {
  Object.entries(data).map(async ([type, object]) => {
    let prevNum = 0;
    Object.entries(object).map(([date, value]) => {
      daily[type as keyof typeof data] = {
        ...daily[type as keyof typeof data],
        [date]: (value as number) - prevNum
      };
      prevNum = value as number;
    });
  });
  return daily;
}
```

puf17640:
what is this type sorcery

buster95:
> something I'm currently using to get the daily data
> 
> ```ts
> function getDaily(data) {
>   Object.entries(data).map(async ([type, object]) => {
>     let prevNum = 0;
>     Object.entries(object).map(([date, value]) => {
>       daily[type as keyof typeof data] = {
>         ...daily[type as keyof typeof data],
>         [date]: (value as number) - prevNum
>       };
>       prevNum = value as number;
>     });
>   });
>   return daily;
> }
> ```

In this PR I was working in that 😥, to incorporate daily cases, daily deaths and daily recovered

puf17640:
i see, but me and ethan both agreed on trying to keep the motto "scrape and serve" without transforming the data too much, since most of the stuff can be done clientside anyway

PR #699, opened 2020-05-19
Hi 

SagarJadhavWebDev:
Can you add critical cases on default page like all world data there are three options deaths , recover, infection please if it's possible add critical option there


ebwinters:
Did you mean to PR this here?

PR #700, opened 2020-05-19
Updating /all json example

MarvelDC:
Is there a reason why this isn't automatic, for the latest data?


puf17640:
because of how the stuff is structured in files

PR #701, opened 2020-05-19
Fix Italy test case

ebwinters:


puf17640:
why did we need to take that out??

puf17640:
oh because it can be below null?

PR #717, opened 2020-05-28
Added live global stats using svg badges

egbakou:
[novelcovid-api-badges](https://github.com/egbakou/novelcovid-api-badges) is a Node js project that displaying coronavirus live global stats through  svg badges.
Demo: https://novelcovid-api-badges.herokuapp.com/


egbakou:
A mistake. will make another PR

PR #730, opened 2020-06-04
V3 update

ebwinters:
- Redirect v2 routes to v3
- Restructure scrapers, tests, and routes folder
- Add Ebola endpoint to v3 routes


config/config.keys.json
```diff
@@ -14,5 +14,6 @@
   "nyt_states": "covidapi:nyt_states",
   "nyt_USA": "covidapi:nyt_usa",
   "apple_all": "covidapi:apple_all",
-  "gov_countries": "covidapi:gov_countries"
+  "gov_countries": "covidapi:gov_countries",
+  "ebola": "covidapi:ebola"
```
puf17640, line 18: I don't see any reason why ebola has `covidapi:ebola` as resource key
we can either have `covidapi:` for covid disease related stuff and `ebola:` for ebola related stuff

routes/v2/apiApple.js
```diff
@@ -0,0 +1,14 @@
+// eslint-disable-next-line new-cap
+const router = require('express').Router();
+
+router.get('/v2/apple/countries/:country?', async (req, res) => {
+	const { country: countryName } = req.params;
```
puf17640, line 5: why would you not just do `const { country } = req.params`

routes/v2/apiApple.js
```diff
@@ -0,0 +1,14 @@
...
+	res.redirect(`/v3/covid-19/apple/countries/${countryName || ''}`);
+});
+
+router.get('/v2/apple/countries/:country/:subregions', async (req, res) => {
+	const { country: countryName, subregions: querySubregions } = req.params;
```
puf17640, line 10: same here

routes/v2/apiGov.js
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line new-cap
+const router = require('express').Router();
+
+router.get('/v2/gov/:country?', async (req, res) => {
+	const { allowNull } = req.query;
+	const { country: countryName } = req.params;
```
puf17640, line 6: why not do `const { country } = req.params`

routes/v2/apiApple.js
ebwinters: the code was already there and it will be deleted when v2 is deprecated anyways

PR #736, opened 2020-06-10
Updating example /all request

MarvelDC:


EliteDaMyth:
not necessary but why not!

PR #745, opened 2020-06-15
refactored 1 param lambdas

puf17640:


puf17640:
okay gotta work on this some more later

puf17640:
something seems to be broken...

puf17640:
image
@ebwinters could you help me find the "circular dependency" cus there should be none..

PR #747, opened 2020-06-17
Added more information from Colombia gov

jsebastianms1:
Added more information from Colombia gov

First time only showed deceased and recovered cases, this shows more information like gender, home isolation, ICU cases, etc.

Colombia has 32 departments and some special districts, so test had to be modified

### DIFF
--- a/scrapers/govScrapers/getColombia.js
+++ b/scrapers/govScrapers/getColombia.js
@@ -1,11 +1,11 @@
 const axios = require('axios');
 const logger = require('../../utils/logger');
 
-const transformData = (data, municipalities) => {
+const transformData = (data, cities) => {
 	const state = 13;
 	const gender = 15;
-	const pos = municipalities ? 11 : 12;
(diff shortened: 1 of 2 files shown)


scrapers/govScrapers/getColombia.js
```diff
@@ -37,7 +47,7 @@ const colombiaData = async () => {
 		return {
 			updated: Date.now(),
 			departments: transformData(colombiaResponse.data, false),
-			cities: transformData(colombiaResponse.data, true)
```
ebwinters, line 40: Don’t call the new field `municipalities`, that would break code that someone may have written to use the `cities`field.

jsebastianms1: You are right, can you please check again? thanks

jsebastianms1:
> I'm probably going to refactor this in a few days since I don't like the way you did `renameKeys` but other than that, looks good to me

I will think on a cleaner way to do this, and submit a new PR, thanks for the comment

PR #753, opened 2020-06-21
closes #751 - added south africa gov scraper 

### LINKED ISSUE #751: [Feature Req] South Africa - Scrape data from Wikipedia
**Is your feature request related to a problem? Please describe.**
Many government websites are too hard to scrape for detailed COVID-19 country data for various reasons (constantly changing formats, data posted as images etc). The official COVID-19 data sources from South Africa is an example of this.

Also, since the current NovelCOVID API endpoints do not provide granular country-level data, I had to go look for another reliable source for detailed South African COVID-19 data.

**Describe the solution you'd like**
https://en.wikipedia.org/wiki/Template:COVID-19_pandemic_data/South_Africa_medical_cases (click "show") lists some COVID-19 data per province within South Africa. Replacing "South_Africa" within the URL with other country names brings up their equivalents. **Note**: I only checked a few countries and found most of them to return something, but in different formats.

I am only really interested in more granular South African data, which that Wikipedia source provides to some degree in what looks to be a scraper friendly format! Wikipedia could be used to scrape data for other countries as well, however, that would require a custom scraper to be coded for each country plus the maintenance work that goes along with it.

All-in-all what I am saying is that I am planning to write code to scrape the South African Wikipedia data myself, but why not write the scraper for this API rather then? It could serve as a data source for the **/gov/SouthAfrica** endpoint?

Let me know your thoughts, if approved I'll go ahead with writing it for this API.

skyzadev:
closes #751  Added a South Africa gov scraper via Wikipedia (government sourced data)

### DIFF
--- a/scrapers/govScrapers/getSouthAfrica.js
+++ b/scrapers/govScrapers/getSouthAfrica.js
@@ -8,16 +8,16 @@ const logger = require('../../utils/logger');
 const southAfricaData = async () => {
 	try {
 		const html = cheerio.load((await axios.get('https://en.wikipedia.org/wiki/Template:COVID-19_pandemic_data/South_Africa_medical_cases')).data);
-		let data = {
+		const data = {
 			national: { timeline: [] },
 			provinces: ['Eastern Cape', 'Free state', 'Gauteng', 'KwaZulu-Natal', 'Limpopo', 'Mpumalanga', 'Northern Cape', 'North West', 'Western Cape']
-				.map(name => { return { name, timeline: [] }; })
+				.map(name => ({ name, timeline: [] }))
 		};
 
-		let tableRows = html("span:contains('COVID-19 confirmed cases in South Africa by province')").closest('table').find('tr');
+		const tableRows = html("span:contains('COVID-19 confirmed cases in South Africa by province')").closest('table').find('tr');
 		tableRows.slice(2, tableRows.length - 2).each((i, tr) => {
-			let tds = html(tr).children();
-			let date = `2020-${tds.eq(0).text()}`.replace(/\n/g, '');
+			const tds = html(tr).children();
+			const date = `2020-${tds.eq(0).text()}`.replace(/\n/g, '');
 			data.national.timeline.push({
 				date: date,
 				cases: { cumulative: getCellData(tds, 15), unallocated: getCellData(tds, 13), new: getCellData(tds, 14) },
@@ -26,8 +26,8 @@ const southAfricaData = async () => {
 				recoveries: { cumulative: getCellData(tds, 20) }
 			});
 
-			data.provinces.forEach((p, i) => {
-				p.timeline.push({ date: date, cases: getCellData(tds, i+4) });
+			data.provinces.forEach((prov, index) => {
+				prov.timeline.push({ date: date, cases: getCellData(tds, index + 4) });
 			});
 		});
 		data.updated = Date.now();
@@ -39,6 +39,6 @@ const southAfricaData = async () => {
 };
 
 function getCellData(tds, index) {
-	return parseInt(tds.eq(index).text().replace(/\n|\D/g, '')) || null
+	return parseInt(tds.eq(index).text().replace(/\n|\D/g, '')) || null;
 }
 module.exports = southAfricaData;
--- a/scrapers/govScrapers/getSouthAfrica.js
+++ b/scrapers/govScrapers/getSouthAfrica.js
@@ -7,126 +7,29 @@ const logger = require('../../utils/logger');
  */
 const southAfricaData = async () => {
 	try {
-		const html = cheerio.load(
-			(
-				await axios.get(
-					'https://en.wikipedia.org/wiki/Template:COVID-19_pandemic_data/South_Africa_medical_cases'
-				)
-			).data
-		);
-
-		var data = {
-			national: {
-				timeline: []
-			},
-			provinces: [
-				{
-					name: 'Eastern Cape',
-					timeline: []
-				},
-				{
-					name: 'Free State',
-					timeline: []
-				},
-				{
-					name: 'Gauteng',
-					timeline: []
-				},
-				{
-					name: 'KwaZulu-Natal',
-					timeline: []
-				},
-				{
-					name: 'Limpopo',
-					timeline: []
-				},
-				{
-					name: 'Mpumalanga',
-					timeline: []
-				},
-				{
-					name: 'Northern Cape',
-					timeline: []
-				},
-				{
-					name: 'North West',
-					timeline: []
-				},
-				{
-					name: 'Western Cape',
-					timeline: []
-				}
-			]
+		const html = cheerio.load((await axios.get('https://en.wikipedia.org/wiki/Template:COVID-19_pandemic_data/South_Africa_medical_cases')).data);
+		let data = {
+			national: { timeline: [] },
+			provinces: ['Eastern Cape', 'Free state', 'Gauteng', 'KwaZulu-Natal', 'Limpopo', 'Mpumalanga', 'Northern Cape', 'North West', 'Western Cape']
+				.map(name => { return { name, timeline: [] }; })
 		};
 
-		var tableRows = html(
-			"span:contains('COVID-19 confirmed cases in South Africa by province')"
-		)
-			.closest('table')
-			.find('tr');
-
+		let tableRows = html("span:contains('COVID-19 confirmed cases in South Africa by province')").closest('table').find('tr');
 		tableRows.slice(2, tableRows.length - 2).each((i, tr) => {
-			var tds = html(tr).children();
-			var date = `2020-${tds.eq(0).text()}`.replace(/\n/g, '');
-
+			let tds = html(tr).children();
+			let date = `2020-${tds.eq(0).text()}`.replace(/\n/g, '');
 			data.national.timeline.push({
 				date: date,
-				cases: {
-					cumulative: parseInt(tds.eq(15).text().replace(/\n|\D/g, '')) || null,
-					unallocated:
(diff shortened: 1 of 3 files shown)


tests/api_gov.spec.js
```diff
@@ -360,4 +361,36 @@ describe('TESTING /v2/gov/new zealand', () => {
...
+		  res.body.national.timeline[90].tests.cumulative.should.equal(761534);
+		  res.body.national.timeline[90].recoveries.cumulative.should.equal(18313);
+		  res.body.national.timeline[90].deaths.cumulative.should.equal(755);
+		  res.body.national.timeline[90].deaths.new.should.equal(50);
+
+		  res.body.provinces.length.should.equal(9);
+		  res.body.provinces[0].name.should.equal('Eastern Cape');
+		  res.body.provinces[0].timeline.length.should.be.at.least(107);
+		  res.body.provinces[0].timeline[85].date.should.equal('2020-05-28');
+		  res.body.provinces[0].timeline[85].cases.should.equal(3306);
```
puf17640, line 389: this seems kinda odd, why is it index 90 and how can you know it will stay there? same for provinces (85).

ebwinters: The data from 90 days into data tracking won't change, since it's in the past, I am assuming. Better to have dynamic tests still though, so we don't have failing tests from a source update.

scrapers/govScrapers/getSouthAfrica.js
```diff
@@ -0,0 +1,138 @@
...
+ * Scrapes more granular South African data from Wikipedia (based on Government sources)
+ */
+const southAfricaData = async () => {
+	try {
+		const html = cheerio.load(
+			(
+				await axios.get(
+					'https://en.wikipedia.org/wiki/Template:COVID-19_pandemic_data/South_Africa_medical_cases'
+				)
+			).data
```
ebwinters, line 15: Can all be one line

scrapers/govScrapers/getSouthAfrica.js
```diff
@@ -0,0 +1,138 @@
...
+		);
+
+		var data = {
```
ebwinters, line 18: use `let` over `var`

scrapers/govScrapers/getSouthAfrica.js
```diff
@@ -0,0 +1,138 @@
...
+				{
+					name: 'Western Cape',
+					timeline: []
+				}
+			]
+		};
+
+		var tableRows = html(
+			"span:contains('COVID-19 confirmed cases in South Africa by province')"
+		)
```
ebwinters, line 64: can be one line

scrapers/govScrapers/getSouthAfrica.js
```diff
@@ -0,0 +1,138 @@
...
+			.closest('table')
+			.find('tr');
+
+		tableRows.slice(2, tableRows.length - 2).each((i, tr) => {
+			var tds = html(tr).children();
+			var date = `2020-${tds.eq(0).text()}`.replace(/\n/g, '');
```
ebwinters, line 70: use `let` over `var`

scrapers/govScrapers/getSouthAfrica.js
```diff
@@ -0,0 +1,138 @@
...
+
+			data.national.timeline.push({
+				date: date,
+				cases: {
+					cumulative: parseInt(tds.eq(15).text().replace(/\n|\D/g, '')) || null,
```
ebwinters, line 75: put this regex: `/\n|\D/g` into a variable to be reused and altered later if need be

scrapers/govScrapers/getSouthAfrica.js
```diff
@@ -0,0 +1,138 @@
...
+				},
+				recoveries: {
+					cumulative: parseInt(tds.eq(20).text().replace(/\n|\D/g, '')) || null
+				}
+			});
+
+			data.provinces[0].timeline.push({
+				date: date,
+				cases: parseInt(tds.eq(4).text().replace(/\n|\D/g, '')) || null
+			});
```
ebwinters, line 95: Duplicate code. Put this into a function, maybe 
```js
addDate(date, index) {
  data.provinces[index].timeline.push({
    date: date,
    cases: parseInt(tds.eq(index+4).text().replace(/\n|\D/g, '')) || null
  });
}
```

scrapers/govScrapers/getSouthAfrica.js
```diff
@@ -0,0 +1,138 @@
...
+
+		var data = {
+			national: {
+				timeline: []
+			},
+			provinces: [
+				{
+					name: 'Eastern Cape',
+					timeline: []
+				},
```
ebwinters, line 26: instead of harcoding all of this, maybe write:
```js
provinces = ['Eastern cape', 'Free state', ...].map(name => { name, timeline: [] });
```

scrapers/govScrapers/getSouthAfrica.js
```diff
@@ -0,0 +1,138 @@
...
+			data.national.timeline.push({
+				date: date,
+				cases: {
+					cumulative: parseInt(tds.eq(15).text().replace(/\n|\D/g, '')) || null,
+					unallocated:
+            parseInt(tds.eq(13).text().replace(/\n|\D/g, '')) || null,
+					new: parseInt(tds.eq(14).text().replace(/\n|\D/g, '')) || null
+				},
+				tests: {
+					cumulative: parseInt(tds.eq(3).text().replace(/\n|\D/g, '')) || null
```
ebwinters, line 81: instead of this line of code so many different times, put it into a function
```js
getCellData(index) {
  return parseInt(tds.eq(index).text().replace(/\n|\D/g, '')) || null
}
```

tests/api_gov.spec.js
skyzadev: Yeah, the purpose here is to check that the correct data gets scraped for that day 28 May 2020. This is essentially just a spot check to catch scenarios where the webpage might change its structure and the column indexes go out of sync causing wrong values to be used within the properties.
I will make it more dynamic by finding the province object where name == 'Eastern Cape' rather than using hardcoded indexes. I'll do the same for the other instances in the test as well.

puf17640:
Looks good to me! @ebwinters take a look at the changes and then merge 👍

tests/api_gov.spec.js
```diff
@@ -360,4 +362,45 @@ describe('TESTING /v2/gov/new zealand', () => {
...
+        .get('/v2/gov/south africa')
+        .end((err, res) => {
+          testBasicProperties(err, res, 200, 'object');
+          res.body.should.have.property('updated');
+          res.body.should.have.property('national');
+		  res.body.should.have.property('provinces');
+
+		  res.body.national.timeline.length.should.be.at.least(107);
+
+		  let dayNational = res.body.national.timeline.find(d => d.date === '2020-06-02');
```
ebwinters, line 378: use `const` since this variable won't change

tests/api_gov.spec.js
```diff
@@ -360,4 +362,45 @@ describe('TESTING /v2/gov/new zealand', () => {
...
+		  dayNational.cases.unallocated.should.equal(6);
+		  dayNational.cases.new.should.equal(1455);
+		  dayNational.tests.cumulative.should.equal(761534);
+		  dayNational.recoveries.cumulative.should.equal(18313);
+		  dayNational.deaths.cumulative.should.equal(755);
+		  dayNational.deaths.new.should.equal(50);
+
+		  res.body.provinces.length.should.equal(9);
+
+		  let province = res.body.provinces.find(p => p.name === 'Eastern Cape');
```
ebwinters, line 391: use `const` since this variable won't change

tests/api_gov.spec.js
```diff
@@ -360,4 +362,45 @@ describe('TESTING /v2/gov/new zealand', () => {
...
+		  should().exist(province);
+		  province.name.should.equal('Eastern Cape');
+		  province.timeline.length.should.be.at.least(107);
+
+		  let dayProvince = province.timeline.find(d => d.date === '2020-05-28');
```
ebwinters, line 396: use `const` since this variable won't change

tests/api_gov.spec.js
puf17640: We can do those too though

puf17640:
I made the changes Ethan proposed, I guess we are ready to merge when the checks have passed.

Nekiono:
Thanks for the changes. After Julian's commit, the code would now look in line with our requirements for API extensions.
What puts this on hold a bit is that it's sourced from Wikipedia, instead of any official website.
Wouldn't it in this case make sense to generalize an endpoint for Wikipedia, instead of making it specific to some countries?

skyzadev:
> Thanks for the changes. After Julian's commit, the code would now look in line with our requirements for API extensions.
> What puts this on hold a bit is that it's sourced from Wikipedia, instead of any official website.
> Wouldn't it in this case make sense to generalize an endpoint for Wikipedia, instead of making it specific to some countries?

The main thing with Wikipedia is that, although several countries have data on there that can be scraped, it would still require a custom scraper for every country as they all differ. So a general Wikipedia endpoint probably won't work as it won't be able to return data in a standard format.

What could work is something like `/wikipedia/countryname` e.g. `/wikipedia/south africa'. Pretty much how you guys have `/gov/countryname` endpoints set up currently.

In my opinion, `/wikipedia/countryname` endpoints can be set up if more Wikipedia scrapers get added to the API. Until then it can live under /gov to keep things tidy?

puf17640:
> In my opinion, /wikipedia/countryname endpoints can be set up if more Wikipedia scrapers get added to the API. Until then it can live under /gov to keep things tidy?

we surely can't do that as this would break applications that use the endpoint in the time it is in /gov
creating /wikipedia/country would be something we could discuss in a new issue in my opinion, and until this is solved, we could keep this open

PR #756, opened 2020-06-25
fixed nz gov and testing timeout

puf17640:
jhu still fucked up


puf17640:
dont expect tests to run bc of jhu, idk if i wanna fix that

puf17640:
i hate linting tbh

ebwinters:
Increasing the timeout doesn't make the test pass... something is wrong with that endpoint

PR #759, opened 2020-06-26
Link Updates

crazyuploader:
While GitHub redirects to the latest URL for the project, I think keeping the links updated makes more sense.


puf17640:
Thanks!

PR #778, opened 2020-07-06
fix #777

### LINKED ISSUE #777: [BUG] Why wordsStandardize returns 'null' but not null ?
**Description**

In [stringUtils](https://github.com/disease-sh/API/blob/master/utils/stringUtils.js) script line `6` :
`const wordsStandardize = (word) => word ? word.trim().toLowerCase().normalize('NFD').replace(/[\u0300-\u036f]/g, '') : 'null';`

**Expected**
`null`

**Found**
`'null'`

puf17640:
closes #777


puf17640:
somehow this completely diverges from the current master branch thing that is going on, imma fix this in my fork and PR then, way easier and won't confuse me as much as this

PR #776, opened 2020-07-05
Add data to Israel Gov endpoint

meirroth:
Removed sickPerLocation request and replaced it with more detailed info in the timeline.
Added active cases by condition.
Added number of patients on ventilators.
Added back the city lock down status ([was removed here](https://github.com/disease-sh/API/commit/0691b8798bcdbe03091a7af84cfa278a03b19a21?diff=unified#diff-3990eaa03a337e3af1a6a50dd859c910R47)).
Added isolated doctors and nurses.
Added hospital stats.

PR #779, opened 2020-07-06
fix for #777

puf17640:


puf17640:
this now includes changes made in #776

ebwinters:
@puf17640 What is the difference in returning `''` versus `'null'`? Were we only checking a string was returned?

PR #780, opened 2020-07-06
Refactor stringUtils.js

MenaiAla:
:heavy_check_mark: Remove meaningless comments.
:heavy_check_mark: Replace `'null'` by` null`.
:heavy_check_mark: Replace` isNaN()` by `Number.isNaN()`.


utils/stringUtils.js
```diff
@@ -1,42 +1,20 @@
...
- */
-const wordsStandardize = (word) => word ? word.trim().toLowerCase().normalize('NFD').replace(/[\u0300-\u036f]/g, '') : 'null';
-
-/**
- * Returns the correct boolean value for a parameter
- * @param 	{string} 	word 	parameter received
- * @returns 	{boolean} 		resulting boolean value
- */
+
+const wordsStandardize = (word) => word ? word.trim().toLowerCase().normalize('NFD').replace(/[\u0300-\u036f]/g, '') : null;
```
puf17640, line 2: you can't just look at one file when changing stuff, if you check the history of that line, you can see i already fixed it like you did, but that caused other problems which would take more work to fix, so I made it return an empty string which, in my opinion, is also perfectly fine

utils/stringUtils.js
```diff
@@ -1,42 +1,20 @@
...
-/**
- * Returns correct value for lastdays param given string input
- * @param 	{string} 	lastdays 	lastdays param
- * @returns 	{number} 			correct lastdays value in integer form
- */
+
 const getLastDays = (lastdays) => {
 	if (lastdays && lastdays === 'all') lastdays = Number.POSITIVE_INFINITY;
-	if (!lastdays || isNaN(lastdays)) lastdays = 30;
+	if (!lastdays || Number.isNaN(lastdays)) lastdays = 30;
```
puf17640, line 17: Number.isNan only checks if the passed value equals to `NaN` while isNan checks whether the passed value is not convertible to a number, so Number.isNan is wrong here

utils/stringUtils.js
puf17640: also check the commit right after and read the message to see why i changed it back to return a string not null

utils/stringUtils.js
MenaiAla: Ah! yes because I did not find tests, I thought that you're checking NaN value.

utils/stringUtils.js
MenaiAla: JSdocs are meaningless comments because they don't provide any additional information. Developers use it to generate API docs in HTML pages. To get clean code only convention can say to developers what the code does. Read Martin C clean codebook.

package-lock.json
```diff
@@ -1,4196 +0,0 @@
-{
```
ebwinters, line 1: Why was this file deleted?

package.json
```diff
@@ -1,55 +1,55 @@
 {
```
ebwinters, line 1: Why was this file changed as part of this PR?

puf17640:
I agree with ethan, I'm not against adding error handling and making the code better but we should discuss this first, would you be so kind and come to our discord to have a chat @MenaiAla ? 👍

MenaiAla:
@ebwinters , it was a mistake by misleading my branches [Stackoverflow questions](https://stackoverflow.com/questions/62765973/how-to-create-branch-with-zero-previous-commits/62767650#62767650)  and I don't know how I made this PR mess, I meant the last changes. But may I know why `tests` specified for `utils` are in the wrong places. In fact, I prefer separating directories by `features` instead of by roles which means each folder has each `tests` folder. This will help us speed up the workflow. Am I wrong?

puf17640:
> @ebwinters , it was a mistake by misleading my branches [Stackoverflow question](https://stackoverflow.com/questions/62765973/how-to-create-branch-with-zero-previous-commits/62767650#62767650) and I don't know how I made this PR mess, I meant the last changes. But may I know why `tests` specified for `utils` are in the wrong places. In fact, I prefer separating directories by `features` instead of by roles which means each folder has each `tests` folder. This will help us speed up the workflow. Am I wrong?

I don't think having a `tests` folder in every part of the project is a good idea, let's have one `tests` folder with subfolders for each part. way more organized and less change from what we have now

ebwinters:
I Agree with @puf17640, let’s keep the way the tests are organized currently

ebwinters:
any update on your work for this PR @MenaiAla?

MenaiAla:
@ebwinters, I close this PR currently until I release my tests.

PR #782, opened 2020-07-07
Refactor transformNull function

MenaiAla:
**Refactoring steps:**

:recycle: Remove JSdocs because they are useless and do not improve the developer experience. The code itself can express what it does. Also, we need to update them for each change.  Reference : [Clean Code](http://nicolecarpenter.github.io/2016/03/08/clean-code-chapter-4-comments.html#:~:text=As%20a%20general%20rule%2C%20comments,the%20code%20is%20not%20clear.)

:arrows_clockwise: Refactor `transformNull` function by : 
*  Handling [uncaught exceptions](https://davidwalsh.name/fix-javascript-errors) for `null` value and `empty object` `({})`.
* Giving default value which is `null` for the `parameter` to avoid `not defined` exception.

:heavy_multiplication_x: Remove duplicated `return`.

**Warning:**

Make sure that all the invocations from outside will be: `convertNullToZero` instead of `transformNull` before your merge.

### DIFF
--- a/utils/countryUtils.js
+++ b/utils/countryUtils.js
@@ -1,12 +1,22 @@
 const stringUtils = require('./stringUtils');
 const countryData = require('./countries');
 
-const transformNull = (object) => {
-	if (typeof object !== 'object') {
-		return object;
+
+const convertNullToZero = (continentOrCountry = null) => {
+
+	const COUNTRY_INFO = 'countryInfo';
+
+	if (typeof continentOrCountry !== "object" || !continentOrCountry) {
+
+		throw new TypeError(`${continentOrCountry}` + "should be a an object with non null value")
+
 	}
-	// eslint-disable-next-line no-return-assign
-	Object.entries(object).forEach((entry) => entry[0] !== 'countryInfo' && (object[entry[0]] = entry[1] === null ? 0 : transformNull(entry[1])));
+	if (Object.keys(continentOrCountry).length === 0) {
+		throw new Error(`${continentOrCountry}` + "should not be an empty object")
+	}
+
+	Object.entries(coc).forEach((entry) => entry[0] !== COUNTRY_INFO && (coc[entry[0]] = entry[1] === null ? 0 : convertNullToZero(entry[1])));
+
 	return object;
 };
 
--- a/utils/countryUtils.js
+++ b/utils/countryUtils.js
@@ -1,11 +1,6 @@
 const stringUtils = require('./stringUtils');
 const countryData = require('./countries');
 
-/**
- * Transform null values to 0 (DEEPLY TRANSFORM)
- * @param {Object} object country/continent object
- * @returns  {Object} 		transformed object
- */
 const transformNull = (object) => {
 	if (typeof object !== 'object') {
 		return object;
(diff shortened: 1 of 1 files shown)


utils/countryUtils.js
```diff
@@ -1,39 +1,32 @@
...
- * @param {Object} object country/continent object
- * @returns  {Object} 		transformed object
- */
-const transformNull = (object) => {
-	if (typeof object !== 'object') {
-		return object;
+
+const convertNullToZero = (continentOrCountry = null) => {
+
+	const COUNTRY_INFO = 'countryInfo';
```
puf17640, line 7: this makes absolutely no sense, why should we have a constant variable for a value we use just once

utils/countryUtils.js
```diff
@@ -1,39 +1,32 @@
...
-/**
- * Transform null values to 0 (DEEPLY TRANSFORM)
- * @param {Object} object country/continent object
- * @returns  {Object} 		transformed object
- */
-const transformNull = (object) => {
-	if (typeof object !== 'object') {
-		return object;
+
+const convertNullToZero = (continentOrCountry = null) => {
```
puf17640, line 5: good naming

utils/countryUtils.js
```diff
@@ -1,39 +1,32 @@
...
+
+	const COUNTRY_INFO = 'countryInfo';
+
+	if (typeof continentOrCountry !== "object" || !continentOrCountry) {
+
+		throw new TypeError(`${continentOrCountry}` + "should be a an object with non null value")
+
+	}
+	if (Object.keys(continentOrCountry).length === 0) {
+		throw new Error(`${continentOrCountry}` + "should not be an empty object")
```
puf17640, line 15: it does not matter if its an empty object, so we dont need to throw an error if it is

ebwinters:
Are you going to respond to these comments @MenaiAla? Otherwise will close PR for now

utils/countryUtils.js
MenaiAla: How could you test the value `countryInfo` in your unit test if you don't have a variable?

utils/countryUtils.js
MenaiAla: Or ``nullToZero` as I see most developers do, but I prefer functions start with a verb to refer the action.

utils/countryUtils.js
MenaiAla: To improve Developer Experience ( DX ).

When you design an API for developers handling errors matter.

May you see it not matter, but with an empty `object` the function will not return an output. I mean the empty object still an empty object.

MenaiAla:
@ebwinters, sorry because we have different Time Zone.

utils/countryUtils.js
puf17640: an empty object is in no way an error that needs to be handled

utils/countryUtils.js
puf17640: just have a look at the tests, I don't see the connection between unittests and having a constant variable for a value that is used exactly once in that file

utils/countryUtils.js
MenaiAla: Yes, you're right it's not an explicit error but it's a silent bug.

utils/countryUtils.js
MenaiAla: Utility functions have not `unit tests` yet!

ebwinters:
we closing this

PR #805, opened 2020-07-30
Update index.ejs

TogamiCookiie:


EliteDaMyth:
@puf17640 we don't actually have an official facebook page. The one was managed by diced's friend. I think we should either remove it, or make a new page. The one which this pr is suggesting isn't the "official" page

EliteDaMyth:
removed facebook. gonna merge.

PR #807, opened 2020-08-01
Fix AgeAndGender section labels - Israel Gov data

meirroth:
Removed “ages” from label, better for i18n.
Corrected segment increments from “0 - 10” to “0 - 9” and so on...
And replaced “ages 90 – 100” with “+90” to match the governments source.
Screenshot from governments website:
Annotation 2020-08-01 192150


meirroth:
Thank you :)
Any updates with the API returning 404 if there is no data?
I saw you have coded something but then [commented it out](https://github.com/disease-sh/API/blob/master/scrapers/covid-19/govScrapers/getGovData.js#L28)

meirroth:
opps…
How do we fix this npm ERR?

PR #809, opened 2020-08-02
Get canada/csv

Collin-St:
COVID-19 table data on the [Canadian government website](https://www.canada.ca/en/public-health/services/diseases/2019-novel-coronavirus-infection.html) is no longer available. Instead, we'll grab the data from their [source csv file](https://health-infobase.canada.ca/src/data/covidLive/covid19.csv). As a result, `getCanada.js` had to be rewritten. 

The csv includes historical COVID-19 data, which may be useful in the future. But for now, I've filtered it down to the current date.

Note: I've added moment.js to the `package.json`. If the package size is an issue, I can find a way around it.

### DIFF
--- a/scrapers/covid-19/govScrapers/getCanada.js
+++ b/scrapers/covid-19/govScrapers/getCanada.js
@@ -1,7 +1,6 @@
 const axios = require('axios');
 const logger = require('../../../utils/logger');
 const csvUtils = require('../../../utils/csvUtils');
-const moment = require('moment');
 
 /**
  * Return array of provinces that match today's date (initial csv is historical)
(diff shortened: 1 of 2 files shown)


EliteDaMyth:
i really hope these tests fucking pass

scrapers/covid-19/govScrapers/getCanada.js
```diff
@@ -1,39 +1,31 @@
...
-			}
-			default: {
-				province[columns[index]] = parseInt(cell.text().replace(/,/g, '')) || null;
-			}
-		}
-	});
-	return province;
+const filterByDate = (csv) => {
+	const date = moment().format('DD-MM-YYYY');
+	return csv.filter(row => row.date === date);
```
ebwinters, line 13: I'd rather not install moment.js for this purpose. It's relatively easy to get today's date in the format of `dd-mm-yyyy`. I'll post a snippet below:
```js
const d = new Date();
return `${d.getDate().padStart(2, '0')}${d.getMonth()}${d.getYear()}`
```

scrapers/covid-19/govScrapers/getCanada.js
```diff
@@ -1,39 +1,31 @@
...
-				break;
-			}
-			default: {
-				province[columns[index]] = parseInt(cell.text().replace(/,/g, '')) || null;
-			}
-		}
-	});
-	return province;
+const filterByDate = (csv) => {
+	const date = moment().format('DD-MM-YYYY');
```
ebwinters, line 12: The entire reason tests are failing, and no data is returned is because you are checking for today's date in the CSV. You need to check YESTERDAY'S date, since the CSV is only up to date as of the day before. Then, tests will work.

PR #821, opened 2020-08-06
Extend mexico

### LINKED ISSUE #820: [Feature Req] getMexico - add data for Active, Suspect, and Negative cases
**Is your feature request related to a problem? Please describe.**
There is extra data to be added to our response.

**Describe the solution you'd like**
I will add additional information to the response for `getMexico` to account for current Active, Suspect, and Negative cases for National and State (if available).

**Additional context**
- Same source, just extra parsing
- This is per request from a community member

Collin-St:
closes #820 

Added additional data from the source: 
- negative, suspect, active cases for state and national
- male, female for national daily cases and deaths
- improved filterByDate to check the source for today's date before falling back to yesterday's date

Known irregularity:
Although the source's National Accumulated Death count is accurate, the National Daily Death count is skewed and does not match figures from other indirect sources such as: 
- https://twitter.com/HLGatell, 
- https://www.worldometers.info/coronavirus/country/mexico/, 
- https://www.nytimes.com/interactive/2020/world/americas/mexico-coronavirus-cases.html
but this is what the government source says. It could be caused by testing inconsistency - confirmed deaths must be associated with a confirmed diagnoses. I will keep an eye on this. 

Sample of updated response body:
```
{
  updated: 1596696661535,
  sourceUpdated: '2020-08-05',
  nationalData: {
    todayCases: { male: 10, female: 2, total: 12 },
    todayDeaths: { male: 4, female: 1, total: 5 },
    casesAccumulated: 456100,
    deathsAccumulated: 49698,
    activeCases: 28962,
    negativeCases: 499915,
    suspectCases: 85845,
    recovered: 304708
  },
  stateData: [
    {
      state: 'aguascalientes',
      confirmed: 4257,
      negative: 8415,
      suspect: 526,
      deaths: 266
    },
    {
      state: 'baja california',
      confirmed: 13915,
      negative: 7950,
      suspect: 1309,
      deaths: 2734
    },
    ...
  ],
  source: 'https://coronavirus.gob.mx/datos/'
}
```


scrapers/covid-19/govScrapers/getMexico.js
```diff
@@ -43,30 +43,48 @@ const formData = {
...
+ */
+// eslint-disable-next-line consistent-return
 const filterByDate = (data) => {
 	const date = new Date();
-	date.setDate(date.getDate() - 1);
-	return data.filter(item => item.date === `${date.getFullYear()}-${(date.getMonth() + 1).toString().padStart(2, '0')}-${date.getDate().toString().padStart(2, '0')}`);
+	const filtered = () => data.filter(item => item.date === `${date.getFullYear()}-${(date.getMonth() + 1).toString().padStart(2, '0')}-${date.getDate().toString().padStart(2, '0')}`);
+	const decrementDate = () => date.setDate(date.getDate() - 1);
+	try {
+		return filtered().length ? filtered() : decrementDate(date) && filtered();
```
ebwinters, line 63: Redundant function calls on this line - `filtered()`
Store this in a variable, and check its length. If the length isn't 0 then return the variable, otherwise decrement date and return the new function call result

scrapers/covid-19/govScrapers/getMexico.js
```diff
@@ -43,30 +43,48 @@ const formData = {
...
+	} catch (err) {
+		logger.err(err);
+	}
```
ebwinters, line 66: I think it would be better to return as well as log an error. This way you can handle errors down the chain of function calls better - perhaps return `null`

scrapers/covid-19/govScrapers/getMexico.js
```diff
@@ -78,25 +98,59 @@ const getState = (res) => {
...
 const mexicoData = async () => {
+	const SOURCE_URL = 'https://coronavirus.gob.mx/datos/Overview/info/getInfo.php';
+
 	try {
-		const nationalCases = getNational((await axios.post('https://coronavirus.gob.mx/datos/Overview/info/getInfo.php', formData.nationalCases)).data);
-		const nationalDeaths = getNational((await axios.post('https://coronavirus.gob.mx/datos/Overview/info/getInfo.php', formData.nationalDeaths)).data);
-		const stateCases = getState((await axios.post('https://coronavirus.gob.mx/datos/Overview/info/getInfo.php', formData.stateCases)).data);
-		const stateDeaths = getState((await axios.post('https://coronavirus.gob.mx/datos/Overview/info/getInfo.php', formData.stateDeaths)).data);
+		const nationalCases = getNational((await axios.post(SOURCE_URL, formData.nationalCases)).data);
+		const nationalDeaths = getNational((await axios.post(SOURCE_URL, formData.nationalDeaths)).data);
```
ebwinters, line 118: since you added a return value from `getNational`, keep in mind this could be set to null is something errors. Maybe a more appropriate error value is the empty list? I only say this because if people are expecting a list you don't want to implode their program by having them check `null.length` or something

Collin-St: you're correct, `getNational` relies on `filterByDate`.

but what would actually happen if `filterByDate` fails is, it would still return an empty array, because Array.filter returns [] if the predicate is never met. this function would only reach the catch if `data` is undefined, which would again bubble up to `mexicoData` where the post request was made.

then the error would bubble up to `mexicoData`, where the catch would again return null. this brings up to `getGovData`. 

see my message on discord about redis

Issue #831, opened 2020-08-14
Re-direct v2 routes to v3 using Express


ebwinters:
We have many high profile users who are unreachable, and are still using v2. Because of this, we will redirect v2 endpoints to their corresponding v3 endpoint, while no longer supporting v2 testing or docs. Here is an example of a route re-direct for the NYT state endpoint:
```js
router.get('/v2/nyt/states/:state', async (req, res) => {
	const { state: queryState } = req.params;
	res.redirect(`/v3/covid19/nyt/states/${queryState || ''}`);
});
```

There are three parts to this issue:
- [ ] Redirects
- [ ] Delete v2 tests
- [ ] Deprecate v2 docs

puf17640:
I will be working on this later on

ebwinters:
We are no longer redirecting, but instead keeping the routes. This issue can be used to deprecate v2 testing and docs though.

PR #832, opened 2020-08-14
# Merge contributions with master

rohitjethoe:
**# Merge contributions with master**
I've changed the code of /v2 to redirect to /v3 API and removed the modules that weren't used anymore because of the redirections.


ebwinters:
Decided against redirecting ... check #831 for update

PR #839, opened 2020-08-16
states endpoint fixed + small jhu and uk scraper change

### LINKED ISSUE #833: [BUG] Cache warning on docker-test
Apple and NYT caches produce warnings on test runs because initially they are not filled

puf17640:
this should also fix #833


puf17640:
so vietnam gov site seems to be having problems...

Issue #835, opened 2020-08-14
[Feature Req] NYT lastDays parameter


ebwinters:
https://disease.sh/v3/covid-19/nyt/counties endpoint takes a very long time to load, and while testing, sometimes causes timeouts. Only returning the last 15-30 days (will need to experiment with what is quicker) by default will be much better performance wise than returning the full data set. This will also speed up testing, while not sacrificing testing data correctness. 
We also have a `lastDays` param for the historical endpoint, so the logic can be copied.

puf17640:
just started working on it, it's a pain in the ass because of how the NYT endpoints work (in particular the counties endpoint)

puf17640:
I might work on this in the next few days, but this is really tedious work ...

ebwinters:
@neo7337 is also working on this

PR #844, opened 2020-08-16
Neo7337/feature835

neo7337:
Raising this PR for the feature reqeust for #835 
The data has been filtered on the base 30 days but is configurable


utils/cache.js
```diff
@@ -12,7 +12,27 @@ exports.currentStatus = {
...
+/**
+ * Parses NYT Counties data
+ * @param 	{string}	lastdays  	How many days to show always take lastest
+ * @returns {Array}				The filtered historical data.
+ */
+exports.nytCounties = (lastdays = 30) => {
+	if (lastdays === 'all') {
+		return this.currentStatus.nytCounties;
+	} else {
+		var priorDate = new Date();
```
ebwinters, line 24: Can you do
```js
const priorDate = new Date().setDate(priorDate.getDate() - lastdays).toISOString().slice(0, 10);
```
It's cleaner and always prefer `const` over `var`.

utils/cache.js
```diff
@@ -12,7 +12,27 @@ exports.currentStatus = {
...
+		priorDate.setDate(priorDate.getDate() - lastdays);
+		priorDate = priorDate.toISOString().slice(0, 10);
+		const priorData = [];
+		this.currentStatus.nytCounties.forEach((data) => {
+			if (data.date >= priorDate) {
+				priorData.push(data);
+			}
+		});
+		return priorData;
```
ebwinters, line 33: Can make this cleaner:
```js
return this.currentStatus.nytCounties.filter((data) => data.date >= priorDate));
```

utils/cache.js
neo7337: I tried this but seems like priorDate is not declared yet so this `priorDate.getDate()` block won't work. So had to keep it the previous way and let it be var as we cannot assign values to a const.

tests/v3/covid-19/api_nyt.spec.js
```diff
@@ -80,6 +80,15 @@ describe('TESTING /v3/covid-19/nyt/counties', () => {
 			});
 	});
 
+	it('/v3/covid-19/nyt/counties lastdays param', (done) => {
+		chai.request(app)
+			.get('/v3/covid-19/nyt/counties?lastdays=1')
+			.end((err, res) => {
+				testBasicProperties(err, res, 200, 'array');
+				done();
+			});
```
ebwinters, line 89: maybe also check if the length of the returned response is less than that of a normal counties length (probably in the 1000s), and also if it is greater than 1 just to make sure the param is working

ebwinters:
@neo7337 I need you to do a couple things:
1. Add docs for this param
2. Add this param to other NYT endpoints, like /county
3. Fix the issue of checking lastDays=1 in prod. You may need to scrape this data more frequently by updating interval, or you may need to adjust to look Date-1-`lastdays` back.

PR #847, opened 2020-08-17
Fix NZ scraper

ebwinters:
Had to remove new cases, since source did


puf17640:
lmao I fixed it too, #848

PR #848, opened 2020-08-17
fixed nz gov and refactored config keys

puf17640:


ebwinters:
Lets merge this PR instead, but neither of us removed that field from testing, so it is still looking for newCases in the tests lol

PR #852, opened 2020-08-23
Issue/#844

neo7337:
- Multiple NYT endpoints have last days param

- Test cases enhanced to cater for the `lastdays=1` issue now checking for `lastdays=2` to be safe

### DIFF
--- a/utils/cache.js
+++ b/utils/cache.js
@@ -11,6 +11,13 @@ exports.currentStatus = {
 	appleData: undefined
 };
 
+const calculatePriorDate = (lastDays) => {
+	var priorDate = new Date();
+	priorDate.setDate(priorDate.getDate() - lastDays);
+	// priorDate = priorDate.toISOString().slice(0, 10);
+	return priorDate.toISOString().slice(0, 10);
+};
+
 // Series of get calls to retrieve current state of cache
 /**
  * Parses NYT Counties data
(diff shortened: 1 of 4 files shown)


Collin-St:
Would it be possible to just map `lastdays=1` to the latest date available in NYT `us-counties.csv` data? Parsing the CSV to json returns an array of rows, so it'd be as simple as popping the last array element and using that date.

utils/cache.js
```diff
@@ -11,6 +11,13 @@ exports.currentStatus = {
 	appleData: undefined
 };
 
+const calculatePriorDate = (lastDays) => {
+	var priorDate = new Date();
+	priorDate.setDate(priorDate.getDate() - lastDays);
+	// priorDate = priorDate.toISOString().slice(0, 10);
```
ebwinters, line 17: delete this comment

utils/cache.js
```diff
@@ -21,13 +28,21 @@ exports.nytCounties = (lastdays = 30) => {
 	if (lastdays === 'all') {
 		return this.currentStatus.nytCounties;
 	} else {
-		var priorDate = new Date();
-		priorDate.setDate(priorDate.getDate() - lastdays);
-		priorDate = priorDate.toISOString().slice(0, 10);
-		return this.currentStatus.nytCounties.filter((data) => data.date >= priorDate);
+		const prior = calculatePriorDate(lastdays);
```
ebwinters, line 31: you can still call this variable `priorDate`, it reads easier.

utils/cache.js
```diff
@@ -21,13 +28,21 @@ exports.nytCounties = (lastdays = 30) => {
...
+	if (lastdays === 'all') {
+		return this.currentStatus.nytStates;
+	} else {
+		const prior = calculatePriorDate(lastdays);
+		return this.currentStatus.nytStates.filter((data) => data.date >= prior);
 	}
 };
-exports.nytStates = () => this.currentStatus.nytStates;
+
+// exports.nytStates = () => this.currentStatus.nytStates;
```
ebwinters, line 45: remove this comment

ebwinters:
@Collin-St can you elaborate on what you mean by this? There isn't really an easy way to map `lastDays` to the last array element, because the way the data is returned is there can be multiple counties with the same name, so you would need to also check there are no duplicate names.
Looks at https://disease.sh/v3/covid-19/nyt/counties/cook

Issue #853, opened 2020-08-25
[BUG] Fix cache hang issue in prod by splitting Redis keys


ebwinters:
Right now is prod is seems that a query to Redis on NYT or Apple data hangs because the keys are literally too big. This is causing the entire cache to become stale, so the fix could be splitting keys up for first NYT and seeing if that fixes the issue, then Apple. 
@Collin-St will work on this

PR #859, opened 2020-09-01
Nyt redis

Collin-St:
A couple things are going on here - redis key splitting for all NYT endpoints, a fix for the lastdays=1 timezone issue, adding lastdays to the `nyt/usa` endpoint, and upping the NYT scraper interval to 6 hours. response bodies are untouched, but response times for all endpoints are reduced greatly.

#### key splitting
- the NYT keys are now [hashes](https://redis.io/topics/data-types) 
- each endpoint has up to `lastdays=30` (this is configurable) cached fully (no filtering for any `lastdays` value necessary for up to 30)
- above `lastdays=30` requires filtering the full dataset for whichever endpoint - but latency is still greatly reduced
- county endpoints up to `lastdays=30` latency is down from ~7-15 seconds to ~20-500 ms (locally). a higher `lastdays` value increases the response time
- county endpoint above `lastdays=30` up to `lastdays=all` latency is down from ~7-15 seconds to ~1.5-2 seconds (locally)
- state endpoints latency is also down by ~50%
- national endpoint was also optimized purely because it allowed cleaner code. it wasn't remotely necessary.

#### lastdays=1 timezone issue
- for ~8 hours each day, any NYT endpoint with the query parameter `lastdays=1` returns an empty array, because the function responsible for filtering by date was calculating the current date in UTC, on each request. this resulted in the server time being out of sync with the latest date available in NYT data
- `lastdays` filtering now works by using the latest date available in each of the *currently cached* NYT datasets as a baseline, and then subtracts from that for any value greater than 1. `lastdays=1` will never return an empty array anymore.
- other endpoints might have the same issue, but will have to be fixed independently

#### adding `lastdays` to the `nyt/usa` endpoint
- this was done hardly as a user convenience, and more because the caching logic made it a no-brainer to add in
- by default, this endpoint will still return the full dataset so as to not disrupt any users


puf17640:
I will have a look at this after work today!

ebwinters:
Looks like there's a lot to look at here. I'll go over it once I have some time, spent the morning getting tests to pass lmao

Collin-St:
closes #853

ebwinters (approved):

PR #870, opened 2020-09-12
Only run scraper for specified tests if 'test-single' command is run

Collin-St:
Just a little change to how the `test-single` script works. Makes local testing a little faster. the `test` script still works the same.


ebwinters (commented):
This is a good idea, can you just pull down the latest from master since I rolled back, and this is going to overwrite that

ebwinters (approved):

PR #888, opened 2020-09-30
Add Malaysia and Australia gov 

phamleduy04:


scrapers/covid-19/govScrapers/getAustralia.js
```diff
@@ -0,0 +1,37 @@
+const nightmare = require('nightmare');
```
phamleduy04, line 1: the package is needed to use because the table does not fully load the table when using Axios.

scrapers/covid-19/govScrapers/getMalaysia.js
```diff
@@ -0,0 +1,44 @@
...
+				break;
+		}
+	});
+	return state;
+};
+
+const malaysiaData = async () => {
+	try {
+		const html = cheerio.load(await browser
+			.goto('https://datastudio.google.com/u/0/reporting/cfca4946-cfc8-4fb4-804f-bb58aa6c35ca/page/1M%27')
```
phamleduy04, line 31: this URL is based from this [Malaysia gov](https://www.dosm.gov.my/v1/index.php?r=column/cone&menu_id=UjJoNk9OalhZWlVHdExiaGF1OW13UT09)

phamleduy04:
i think nightmare has some problems with docker, I will try to change nightmare to pupperteer

Collin-St:
> i think nightmare has some problems with docker, I will try to change nightmare to pupperteer

Nah avoid puppeteer. The package is huge, and headless browsers use a lot of CPU in general. Is the raw data available anywhere statically so you don't need to spin up a browser?

phamleduy04:
they use google datastudio and microsoft thing, I'm not sure if I can get that....

phamleduy04:
i found this on dev tools 
image
but idk how to parse that data

ebwinters:
Lets try to not use a new package in the project when we can -- especially for a gov scraping file that realistically will not be used by more than one or two people.

EliteDaMyth:
> i found this on dev tools
> image
> but idk how to parse that data

It is a simple get request to the url, and the response is in json format, it isn't hard parsing it, try again, join our discord and ask for help there if you are stuck and I'll help you.

puf17640:
please join our discord so we can give you directions on how to parse this 👍

phamleduy04:
I'm already in the discord

Collin-St:
btw, don't commit your `package-lock.json`

PR #892, opened 2020-10-04
Remove mexico gov

ebwinters:
image

Tired of the source stalling our pipeline


puf17640:
@ebwinters please don't remove scrapers without taking them out of the update-checker 😄

ebwinters:
yeah forgot about that

PR #918, opened 2020-10-15
Add Indonesia Covid-19 govScraper

reinhart1010:
Hello, I would like to add another govScraper for Indonesia, which is taken from the following sources. I also have updated the README file to include this new source (as https://covid19.go.id) as a credit.

```
https://data.covid19.go.id/public/api/data.json
https://data.covid19.go.id/public/api/update.json
https://data.covid19.go.id/public/api/prov.json
```
Here I will leave the responses as-is from the server, noting that they mostly include terms in Indonesian language, such as `jumlah_kasus` as the total case for a given date or province.


ebwinters:
You need to add tests for this under the v2 and v3 routes

ebwinters:
Gonna look @ this tomorrow

ebwinters (approved):
Nice work here! @reinhart1010 

I will say though, gov endpoints tend to be very unstable. If the tests start failing, I'm just going to disable the endpoint until you can fix it. We don't pay much attention to these endpoints as they are not used very much.

reinhart1010:
Thank you.

As far as I know, the structure of this website has been stable since April 2020, and the database seems to be served from ArcGIS. So, the format of the data should also be stable for the next few months, unless if they are changing the contents of the ArcGIS map.

PR #941, opened 2020-11-14
Add a comment to explain case 15 inside mapRows in Worldometer Scraper code file

bhaidar:
While going through the code I got confused with the `case 15` inside the `mapRows()` function in the Worldometer Scraper file. 

When debugging the Worldometer page using the Dev Tools, I was always getting the number of cells per row was 15, hence index 15 was out of bound.

However, the Worldometer page has a `Columns` select element where in my case, I had `New Recovered` unselected. 

Seems `Cheerio` loads all cells even those unselected inside the `Columns` select element.


puf17640 (approved):
lgtm

Issue #938, opened 2020-11-11
[BUG] Recovered values for Canada


SuperMnr:
**Describe the bug**

In historical statistics, all the recovered values for Canada are at 0.

**To Reproduce**

- Live URL for endpoint : https://disease.sh/v3/covid-19/historical/Canada?lastdays=30

ebwinters:
This was by design, due to how JHUCSSSE format's their data for Canada only. You can see here, we are getting cases data from [this](https://github.com/CSSEGISandData/COVID-19/blob/master/csse_covid_19_data/csse_covid_19_time_series/time_series_covid19_confirmed_global.csv) file: 
<img width="1191" alt="image" src="https://user-images.githubusercontent.com/4297028/99079192-96781280-258d-11eb-8e55-496c4fdaa874.png">
Notice all of Canada's provinces are present, but there is no row with general Canada statistics.

Now [here](https://github.com/CSSEGISandData/COVID-19/blob/master/csse_covid_19_data/csse_covid_19_time_series/time_series_covid19_recovered_global.csv) is the source of recovered data: Notice there are no Rows for the provinces, but one for all of Canada:
<img width="1051" alt="image" src="https://user-images.githubusercontent.com/4297028/99079560-1ef6b300-258e-11eb-91eb-8c430fa4e0cd.png">

We did not program an edge case for this, since it just wasn't worth the time. JHU has province data in the recovered file for all other countries with provinces (like China), so Canda is the only real issue. If you want to fix it, see the `historical.js` file in the `scrapers` folder, and if you need help figuring out where to start, we can make a thread under this issue. Otherwise, if you don't want to fix it, I don't think it's high priority enough for else to either

ebwinters:
@SuperMnr do you want to work on this or nah?

nibble0101:
@ebwinters . It appears nobody else is interested in fixing this issue. Can I work on it?

puf17640:
> @ebwinters  and @puf17640. It appears nobody else is interested in fixing this issue. Can I work on it?

Sure! I'll assign you, are you a member of our discord server? If not, you might wanna join so you can ask stuff if something isn't clear. :+1:

nibble0101:
> and if you need help figuring out where to start, we can make a thread under this issue

@ebwinters  @puf17640  Can you guide me on how to proceed here?

ebwinters:
[...]

PR #946, opened 2020-11-21
Handled edge case for Canada

nibble0101:
This is a first draft of the PR which might fix issue #938. Testing it on my setup seems to indicate the fix is not working or it could be due to lag in data scraping. I am really not sure.  Any feedback is welcome.

### DIFF
--- a/scrapers/covid-19/historical.js
+++ b/scrapers/covid-19/historical.js
@@ -162,14 +162,8 @@ const getHistoricalCountryDataV2 = (data, query, province = null, lastdays = 30)
 				}
 			}
 		}
-	} else {
-		for (let i = 0; i < data.length; i++) {
-			if (data[i].country.toLowerCase() === 'canada' && data[i].province.toLowerCase() === 'recovered-aggregate') {
-				data.splice(i, 1);
-				break;
-			}
-		}
 	}
+
 	// filter to either specific province, or provinces to sum country over
 	const countryData = data.filter(item => {
 		const deepMatch = () => stringUtils.wordsStandardize(item.country) === standardizedCountryName
--- a/scrapers/covid-19/historical.js
+++ b/scrapers/covid-19/historical.js
@@ -77,6 +77,29 @@ const historicalV2 = async (keys, redis) => {
 			: parsedAtIndex['Province/State'].toLowerCase();
 		return newElement;
 	});
+	/* This is an edge case to handle recovered data for Canada.
+	   Since JHU dataset doesn't have recovered data for Canada
+	   disaggregated by Province/Region, the aggregate data is
+	   lost when recovered data is formatted in formatRecoveredData
+	   function. This  will create a non-existent Province/Region
+	   called recovered-aggregate with cases and deaths set to 0
+	   and recovered is set to aggregate recovered data. This data
+	   is pulled only if historical data for Canada is requested.
+	*/
+	for (let i = 0; i < result.length; i++) {
+		if (result[i].country.toLocaleLowerCase() === 'canada') {
+			const data = { country: result[i].country, countryInfo: { ...result[i].countryInfo }, province: 'recovered-aggregate', timeline: { cases: {}, deaths: {}, recovered: {} } };
+			const recoveredCanada = parsedRecovered.find(country => country['Country/Region'] === 'Canada' && country['Province/State'] === '');
+			if (!recoveredCanada) break;
+			Object.keys(data.timeline.recovered).forEach(date => {
+				data.timeline.recovered[date] = parseInt(recoveredCanada[date]);
+				data.timeline.deaths[date] = 0;
+				data.timeline.cases[date] = 0;
+			});
+			result.splice(i, 0, data);
+			break;
+		}
+	}
 
 	redis.set(keys.historical_v2, JSON.stringify(result));
 	logger.info(`Updated JHU CSSE Historical: ${result.length} locations`);
@@ -90,6 +113,13 @@ const historicalV2 = async (keys, redis) => {
  */
 const getHistoricalDataV2 = (data, lastdays = 30) => {
 	lastdays = stringUtils.getLastDays(lastdays);
+	// Removing data inserted for edge case of Canada
+	for (let i = 0; i < data.length; i++) {
+		if (data[i].country.toLowerCase() === 'canada' && data[i].province.toLowerCase() === 'recovered-aggregate') {
+			data.splice(i, 1);
+			break;
+		}
+	}
 	return data.map(country => {
 		delete country.countryInfo;
 		const cases = {}, deaths = {}, recovered = {};
@@ -116,6 +146,30 @@ const getHistoricalCountryDataV2 = (data, query, province = null, lastdays = 30)
 	lastdays = stringUtils.getLastDays(lastdays);
 	const countryInfo = nameUtils.getCountryData(query);
 	const standardizedCountryName = stringUtils.wordsStandardize(countryInfo.country ? countryInfo.country : query);
+	/*
+	  If data required is for a Province/Region of
+	  Canada or data required is for a country other
+	  than Canada, remove the edge case data inserted and
+	  and keep it if general historial data for Canada
+	  is required.
+	*/
+	if (standardizedCountryName.toLowerCase() === 'canada') {
+		if (province !== null) {
+			for (let i = 0; i < data.length; i++) {
+				if (data[i].country.toLowerCase() === 'canada' && data[i].province.toLowerCase() === 'recovered-aggregate') {
+					data.splice(i, 1);
+					break;
+				}
+			}
+		}
+	} else {
+		for (let i = 0; i < data.length; i++) {
+			if (data[i].country.toLowerCase() === 'canada' && data[i].province.toLowerCase() === 'recovered-aggregate') {
+				data.splice(i, 1);
+				break;
+			}
+		}
+	}
 	// filter to either specific province, or provinces to sum country over
 	const countryData = data.filter(item => {
 		const deepMatch = () => stringUtils.wordsStandardize(item.country) === standardizedCountryName
@@ -147,6 +201,20 @@ const getHistoricalCountryDataV2 = (data, query, province = null, lastdays = 30)
 			});
 		});
 	});
+	/*
+	This is to remove imaginary Region/Province inserted
+	as edge case if data for Canada is requested without a
+	province. It prevents returning the imaginary Region/Province
+	among the list of Regions/Provinces.
+	*/
+	if (standardizedCountryName.toLowerCase() === 'canada' && province === null) {
+		for (let i = 0; i < provinces.length; i++) {
+			if (provinces[i] === 'recovered-aggregate') {
+				provinces.splice(i, 1);
+				break;
+			}
+		}
+	}
 	return {
 		country: countryData[0].country || standardizedCountryName,
 		province: province ? countryData[0].province || province : provinces,
--- a/scrapers/covid-19/historical.js
+++ b/scrapers/covid-19/historical.js
@@ -91,7 +91,7 @@ const historicalV2 = async (keys, redis) => {
 			const data = { country: result[i].country, countryInfo: { ...result[i].countryInfo }, province: 'recovered-aggregate', timeline: { cases: {}, deaths: {}, recovered: {} } };
 			const recoveredCanada = parsedRecovered.find(country => country['Country/Region'] === 'Canada' && country['Province/State'] === '');
 			if (!recoveredCanada) break;
-			Object.keys(data.timeline.recovered).forEach(date => {
+			Object.keys(result[i].timeline.recovered).forEach(date => {
 				data.timeline.recovered[date] = parseInt(recoveredCanada[date]);
 				data.timeline.deaths[date] = 0;
 				data.timeline.cases[date] = 0;
--- a/scrapers/covid-19/historical.js
+++ b/scrapers/covid-19/historical.js
@@ -53,6 +53,14 @@ const historicalV2 = async (keys, redis) => {
 	const formatedRecovered = formatRecoveredData(parsedCases, parsedRecovered);
 	// dates key for timeline
 	const timelineKey = Object.keys(parsedCases[0]).splice(timelineIndex);
+
+	// For handling edge case for Canada's recovered aggregate data
+	const recoveredAggregateCanada = parsedRecovered.find(country => country['Country/Region'] === 'Canada' && country['Province/State'] === '');
+	const recoveredAggregateCanadaTimeline = Object.values(recoveredAggregateCanada).splice(timelineIndex);
+	const newElementCanada = {
+		country: '', countryInfo: {}, province: 'recovered-aggregate', timeline: { cases: {}, deaths: {}, recovered: {} }
+	};
+
 	// format csv data to response
 	const result = parsedCases.map((_, index) => {
 		const newElement = {
@@ -66,6 +74,20 @@ const historicalV2 = async (keys, redis) => {
 			newElement.timeline.cases[timelineKey[i]] = parseInt(cases[i]);
 			newElement.timeline.deaths[timelineKey[i]] = parseInt(deaths[i]);
 			newElement.timeline.recovered[timelineKey[i]] = parseInt(recovered[i] || 0);
+
+			/*
+			This sets cases and deaths of the timeline field to 0 and populates
+			recovered field with aggregate recovered data for Canada. This ensures
+			conformity with other real provinces of Canada. This is a fake province
+			which is holding aggregate recovered data for Canada. This has to be done
+			here because the original data values are strings which need to be parsed
+			to integers.
+			*/
+			if (index === 0) {
+				newElementCanada.timeline.cases[timelineKey[i]] = 0;
+				newElementCanada.timeline.deaths[timelineKey[i]] = 0;
+				newElementCanada.timeline.recovered[timelineKey[i]] = parseInt(recoveredAggregateCanadaTimeline[i]);
+			}
 		}
 
 		// add country info to support iso2/3 queries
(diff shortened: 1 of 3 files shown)


scrapers/covid-19/historical.js
```diff
@@ -77,6 +77,29 @@ const historicalV2 = async (keys, redis) => {
...
+	/* This is an edge case to handle recovered data for Canada.
+	   Since JHU dataset doesn't have recovered data for Canada
+	   disaggregated by Province/Region, the aggregate data is
+	   lost when recovered data is formatted in formatRecoveredData
+	   function. This  will create a non-existent Province/Region
+	   called recovered-aggregate with cases and deaths set to 0
+	   and recovered is set to aggregate recovered data. This data
+	   is pulled only if historical data for Canada is requested.
+	*/
+	for (let i = 0; i < result.length; i++) {
```
ebwinters, line 89: Instead of looping over this big ass data object, can we do something in the loop above that picks it out of the group. Like before we `return`, we have some variable that we can set called `canadaData`.
```js
newElement.country = countryData.country || parsedAtIndex['Country/Region'];
newElement.countryInfo = countryData;
newElement.province = parsedAtIndex['Province/State'] === '' ? null
    : parsedAtIndex['Province/State'].toLowerCase();
// Like this
if (newElement.country.toLocaleLowerCase() === 'canada' && country['Province/State'] === '') `canadaData` = newElement
```

scrapers/covid-19/historical.js
```diff
@@ -77,6 +77,29 @@ const historicalV2 = async (keys, redis) => {
...
+		if (result[i].country.toLocaleLowerCase() === 'canada') {
+			const data = { country: result[i].country, countryInfo: { ...result[i].countryInfo }, province: 'recovered-aggregate', timeline: { cases: {}, deaths: {}, recovered: {} } };
```
ebwinters, line 91: If you do the above comment this will already be here for you to make a deep copy of the `canadaData` object
```js
data = JSON.parse(JSON.stringify(canadaData);
```

scrapers/covid-19/historical.js
```diff
@@ -77,6 +77,29 @@ const historicalV2 = async (keys, redis) => {
...
+			const recoveredCanada = parsedRecovered.find(country => country['Country/Region'] === 'Canada' && country['Province/State'] === '');
+			if (!recoveredCanada) break;
+			Object.keys(data.timeline.recovered).forEach(date => {
+				data.timeline.recovered[date] = parseInt(recoveredCanada[date]);
+				data.timeline.deaths[date] = 0;
+				data.timeline.cases[date] = 0;
+			});
+			result.splice(i, 0, data);
```
ebwinters, line 99: What does this do?

scrapers/covid-19/historical.js
```diff
@@ -90,6 +113,13 @@ const historicalV2 = async (keys, redis) => {
  */
 const getHistoricalDataV2 = (data, lastdays = 30) => {
 	lastdays = stringUtils.getLastDays(lastdays);
+	// Removing data inserted for edge case of Canada
+	for (let i = 0; i < data.length; i++) {
+		if (data[i].country.toLowerCase() === 'canada' && data[i].province.toLowerCase() === 'recovered-aggregate') {
+			data.splice(i, 1);
+			break;
+		}
+	}
```
ebwinters, line 122: Why are we removing it?

scrapers/covid-19/historical.js
```diff
@@ -116,6 +146,30 @@ const getHistoricalCountryDataV2 = (data, query, province = null, lastdays = 30)
...
+			}
+		}
+	} else {
+		for (let i = 0; i < data.length; i++) {
+			if (data[i].country.toLowerCase() === 'canada' && data[i].province.toLowerCase() === 'recovered-aggregate') {
+				data.splice(i, 1);
+				break;
+			}
+		}
+	}
```
ebwinters, line 172: Again, why are we removing here, we won't be using it anyways

scrapers/covid-19/historical.js
```diff
@@ -147,6 +201,20 @@ const getHistoricalCountryDataV2 = (data, query, province = null, lastdays = 30)
...
+	among the list of Regions/Provinces.
+	*/
+	if (standardizedCountryName.toLowerCase() === 'canada' && province === null) {
+		for (let i = 0; i < provinces.length; i++) {
+			if (provinces[i] === 'recovered-aggregate') {
+				provinces.splice(i, 1);
+				break;
+			}
+		}
+	}
```
ebwinters, line 217: Consider moving this chunk of code to a function, you use it many times

scrapers/covid-19/historical.js
```diff
@@ -147,6 +201,20 @@ const getHistoricalCountryDataV2 = (data, query, province = null, lastdays = 30)
 			});
 		});
 	});
+	/*
```
ebwinters, line 204: Something else to consider: We don't want to return the fake povince with the actual data, especially when it is used. In this list of provinces returned
image
it shouldn't be there

ebwinters (commented):
Left some comments and questions. Going to test the code out locally later

scrapers/covid-19/historical.js
nibble0101: Noted.

scrapers/covid-19/historical.js
nibble0101: Essentially I am inserting a fake `Region/Province` for Canada in the dataset which will hold cummulative aggregate data. All other `Regions/Provinces` have  `deaths`, `cases` and `recovered` fields in the `timeline`.  I am doing the same for the fake `Region/Province` I am inserting. The only difference is that  this fake `Region/Province` is supposed to hold aggregate recovered data for all provinces of Canada. I have set the `deaths` and `recovered` fields to 0 since that data will be extracted from the dataset once requested.

scrapers/covid-19/historical.js
nibble0101: If I understood correctly, ` getHistoricalDataV2` is invoked after hitting the endpoints `/v3/covid-19/historical` and `/v3/covid-19/historical/all` which returns all data for the last 30 days. Since data for countries with `Provinces/Regions` are disaggregated by `Provinces/Regions` then the fake province inserted primarily for holding cumulative aggregate recovered data for Canada will also be included in the returned data if it is not removed.  I hope  `data` is a deep copy of what has been scraped.

scrapers/covid-19/historical.js
nibble0101: TBH I don't think the `else` block is necessary if the requested data is not for `Canada`. I don't know why I included it there.

scrapers/covid-19/historical.js
nibble0101: I have taken care of that with the line of code below
```js
if (standardizedCountryName.toLowerCase() === 'canada' && province === null) {
		for (let i = 0; i < provinces.length; i++) {
			if (provinces[i] === 'recovered-aggregate') {
				provinces.splice(i, 1);
				break;
			}
		}
	}
```

nibble0101:
@ebwinters  I have left some comments in the new commits I have just made. At least it works on my end. I have just realized the global aggregate recovered data was not including recovered data for Canada. This PR seems to sort that as well. You can also test it on your end and let me know what you think.

ebwinters:
@nibble0101 sure thing, thanks for taking the time. It may take me a few days since to get a full review done, just heads up

scrapers/covid-19/historical.js
```diff
@@ -66,6 +74,20 @@ const historicalV2 = async (keys, redis) => {
...
+
+			/*
+			This sets cases and deaths of the timeline field to 0 and populates
+			recovered field with aggregate recovered data for Canada. This ensures
+			conformity with other real provinces of Canada. This is a fake province
+			which is holding aggregate recovered data for Canada. This has to be done
+			here because the original data values are strings which need to be parsed
+			to integers.
+			*/
+			if (index === 0) {
```
ebwinters, line 86: Why if `index === 0`? This seems out of place here, if you're just trying to do this once, do it where you initialize `timeline: { cases: {}, deaths: {}, recovered: {}` ie
```js
for (let i = 0; i < cases[0].length; i++) { ...
```

scrapers/covid-19/historical.js
```diff
@@ -90,6 +137,13 @@ const historicalV2 = async (keys, redis) => {
  */
 const getHistoricalDataV2 = (data, lastdays = 30) => {
 	lastdays = stringUtils.getLastDays(lastdays);
+	// Removing data inserted for edge case of Canada
+	for (let i = 0; i < data.length; i++) {
+		if (data[i].country.toLowerCase() === 'canada' && data[i].province.toLowerCase() === 'recovered-aggregate') {
+			data.splice(i, 1);
+			break;
+		}
```
ebwinters, line 145: this code is repeated, can you put it in a function called `removeReoveredAggregateCanada` or something? Would just clean up the code a bit

ebwinters (commented):
Much better job on this iteration. Consider the comments I left though, and also, very important, add some tests to make sure we are getting data for this fake province, and that when we request other countries, the fake province is correctly removed.

scrapers/covid-19/historical.js
nibble0101: > Why if index === 0? This seems out of place here, if you're just trying to do this once, do it where you initialize timeline: { cases: {}, deaths: {}, recovered: {} ie

You are right. It is `index === 0` because  I want to do it once. The aggregate recovered data  have string values. Essentially I want to take advantage of the loop to parse the string values into integers as I explained in the preceding comment.  If you want it removed, then I can take outside the callback.

scrapers/covid-19/historical.js
nibble0101: Sure. Will do.

scrapers/covid-19/historical.js
ebwinters: Ok cool, lets move it out

ebwinters:
Looking good @nibble0101 although it still feels hacky, there really isn't any way around this than custom code. 

Going to test out locally, but add some tests for this when you get a chance

nibble0101:
> Looking good @nibble0101 although it still feels hacky, there really isn't any way around this than custom code.
> 
> Going to test out locally, but add some tests for this when you get a chance
> 
> EDIT: Tested locally, looking good

Sure. I will work on it before next week.

nibble0101:
I have failed to test on my end. Running `npm run test` always ends in an error.  I have noticed it is failing here as well.

tests/v3/covid-19/api_historical.spec.js
```diff
@@ -148,6 +148,38 @@ describe('TESTING /v3/covid-19/historical', () => {
 			});
 	});
 
+	// Testing fake province added specifically for handling recovered aggregate for Canada
+	it('/v3/covid-19/historical', (done) => {
+		chai.request(app)
+			.get('/v3/covid-19/historical')
+			.end((err, res) => {
+				testBasicProperties(err, res, 200, 'array');
+				res.body.length.should.equal(271);
```
ebwinters, line 157: don't these types of values, the source could add/remove countries to their collection at will and it will break this test. If you want to make sure the object is full, check >= 1

ebwinters (commented):
tests look good except just the one that isn't real good lol. nice work

tests/v3/covid-19/api_historical.spec.js
nibble0101: Will checking `>=1` serve our purpose here? Essentially my intention was to check if the fake province has been removed from the list of countries returned. Though it looks off, I am thinking about something like:
```js
res.body.forEach(country => { country.province.should.not.be('recovered-aggregate')});
```
Still new to chai. I am not sure it is even valid to do something like that.

ebwinters: you already check that here though: 
```js
			.get('/v3/covid-19/historical/canada/recovered-aggregate')
			.end((err, res) => {
				testBasicProperties(err, res, 404, 'object');
				res.body.should.have.property('message');
				done();
			});
```
Also yeah that is valid I believe.

ebwinters: > I have failed to test on my end. Running `npm run test` always ends in an error. It failed even before adding this particular tests therefore I don't believe it is because of them. I have noticed it is failing here as well. Below is the error message I am getting.
> 
> ```js
>  1) "before all" hook in "{root}"
> 
>   0 passing (1m)
>   1 failing
> 
>   1) "before all" hook in "{root}":
>      Error: Timeout of 70000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
>       at listOnTimeout (node:internal/timers:555:17)
>       at processTimers (node:internal/timers:498:7)
> ```

@nibble0101 I pushed some code to your branch that removes some useless scrapers, basically they take too long and we don't even support the endpoints anymore. Testing should work now, but also to test your own we have a script for that:
`npm run test-single tests/v3/covid-19/api_historical.spec.js`

ebwinters:
right now you're failing one test, the other failures are transient (I'm considering disabling the therapeutics tests too lol)

```
1) TESTING /v3/covid-19/historical
covid-api-app-test |        /v3/covid-19/historical/:country/:province fake province for Canada:
covid-api-app-test | 
covid-api-app-test |       Uncaught AssertionError: expected { Object (_events, _eventsCount, ...) } to have status code 404 but got 200
covid-api-app-test |       + expected - actual
covid-api-app-test | 
covid-api-app-test |       -200
covid-api-app-test |       +404
covid-api-app-test |       
covid-api-app-test |       at testBasicProperties (tests/testingFunctions.js:14:18)
covid-api-app-test |       at /home/container/tests/v3/covid-19/api_historical.spec.js:177:5
covid-api-app-test |       at Test.Request.callback (node_modules/superagent/lib/node/index.js:716:12)
covid-api-app-test |       at /home/container/node_modules/superagent/lib/node/index.js:916:18
covid-api-app-test |       at IncomingMessage.<anonymous> (node_modules/superagent/lib/node/parsers/json.js:19:7)
covid-api-app-test |       at IncomingMessage.emit (node:events:388:22)
covid-api-app-test |       at IncomingMessage.EventEmitter.emit (node:domain:470:12)
covid-api-app-test |       at endReadableNT (node:internal/streams/readable:1294:12)
covid-api-app-test |       at processTicksAndRejections (node:internal/process/task_queues:80:21)
```

nibble0101:
> right now you're failing one test, the other failures are transient (I'm considering disabling the therapeutics tests too lol)

That particular test is failing because of the following lines of code.
```js
else if (provinces.length > 0) {
		// provinces for one country
		countryData = provinces.map((prov) =>
			scraper.historical.getHistoricalCountryDataV2(
				data,
				countries[0],
				prov.trim(),
				lastdays
			) || { message: 'Country not found or doesn\'t have any historical data' }
		);
	}
```
and 
```js
if (countryData) {
		res.send(countryData.length === 1 ? countryData[0] : countryData);
} else {
		res.status(404).send({ message: 'Country not found or doesn\'t have any historical data' });
}
```
Essentially requesting data for a non-existent province of any country with provinces doesn't return a `404` status code as I had thought. `scraper.historical.getHistoricalCountryDataV2(data, countries[0], prov.trim(), lastdays)` returns `null` and the value of `countryData` would be `[{ message: 'Country not found or doesn\'t have any historical data' }]`. The response will then be `{ message: 'Country not found or doesn\'t have any historical data' }` with status code of `200`.  Shouldn't the status code be `404`? Anyway let me fix the test to handle a status code of `200`.

tests/v3/covid-19/api_historical.spec.js
nibble0101: > you already check that here though:
> 
> ```js
> 			.get('/v3/covid-19/historical/canada/recovered-aggregate')
> 			.end((err, res) => {
> 				testBasicProperties(err, res, 404, 'object');
> 				res.body.should.have.property('message');
> 				done();
> 			});
> ```
> 
> Also yeah that is valid I believe.

Essentially there are two endpoints where the fake data has to be removed. In the test below I am trying to ensure the fake province has been removed if data for all countries has been requested.
```js
.get('/v3/covid-19/historical')
			.end((err, res) => {
				testBasicProperties(err, res, 200, 'array');
				res.body.length.should.equal(271);
```
Well what do you suggest? Should I remove the test case?

ebwinters:
Pushed some code to your branch that fixes the 404 issue, that was my bad from like March or whenever we were scrambling to build this historical data functionality. It was a known issue just never bothered fixing

tests/v3/covid-19/api_historical.spec.js
```diff
@@ -109,15 +109,15 @@ describe('TESTING /v3/covid-19/historical', () => {
...
+				testBasicProperties(err, res, 404, 'array');
 				res.body[0].should.have.property('message');
 				done();
 			});
 	});
 
 	it('/v3/covid-19/historical/ correct province name list', (done) => {
 		chai.request(app)
-			.get('/v3/covid-19/historical/156/bejing,hubei')
+			.get('/v3/covid-19/historical/156/bejing,hebei')
```
nibble0101: `bejing` is not a province in China (I changed it to `beijing` in my latest commit). In the endpoint, you are checking only the first element in the array. If it is not `null`, the response status code will be `200` but in this case because it is `null`, the `else` clause will be evaluated resulting in response status code of `404` though the second province has data. As a result the above test fails. I suggest you only return status code of 404 if all the provinces are null. Something like:
```js
if (countryData && (countryData.some(data => data !== null)) {
		res.send(countryData.length === 1 ? countryData[0] : countryData);
	} else 
```

nibble0101 (commented):
@ebwinters  Some observations here.

tests/v3/covid-19/api_historical.spec.js
ebwinters: I see the issue, good catch. Do you want to fix since you already have the code here?

Basically just need this to be fixed and then we can merge your feature

nibble0101: Well in the process of fixing the above code, I have also noticed that if  data for a single country is requested without a province or region, the data returned is not an array. From:
```js
countryData = scraper.historical.getHistoricalCountryDataV2(
			data,
			query,
			province,
			lastdays
		);
```
Essentially `countryData` is an object with the shape:
```js
{
  "country": "string",
  "province": [
    "string"
  ],
  "timeline": {
    "cases": {
      "date": 0
    },
    "deaths": {
      "date": 0
    },
    "recovered": {
      "date": 0
    }
  }
}
```
But then we are using the length of `countryData` in `res.send(countryData.length === 1 ? countryData[0] : countryData);` All along this wasn't noticed, I guess, is because `{}.length === 1` evaluates to `false` since `{}.length` is `undefined` hence returning `countryData`. So to fix the above I have decided to use `countries.map` in the `else` clause where `countryData` is assigned a value so that it will always be an array. You can find it in my latest commit.

ebwinters: Good find, I think this is ready to merge yeah? 😄

nibble0101: I believe so.

ebwinters (approved):

ebwinters:
Nice contribution here @nibble0101 

Make sure you're in our discord, you did good work it's always good to see community members getting acclimated to the code and making solid contributions!

nibble0101:
Thanks. I am a member of the discord server.

PR #955, opened 2020-12-10
Fix IL Gov endpoint & new data

meirroth:
### Summery
1. Added Israel Gov to sources on README.
2. Removed ID's from params requests as they are not relevant anymore, IDs at the source have changed.
3. Added data on deaths, ventilators and critical by age and gender. And restructured the age and gender endpoint. What is the policy with making changes to endpoints?
4. Source changed for health personnel data and verified cases was removed. I set it to NULL.
5. City data has been updated on the Gov's dash, and now shows information regarding the new Traffic Light Roadmap that was set. (About it here: https://www.gov.il/en/Departments/news/30082020_05)
    I have added this new data to the API endpoint trafficLightMap (Cant come up with a better name...), it has its own last update value.
    I am not sure if the other endpoint (cityData) should remain, it seems to be updated, but it is not used anymore on the Gov's website.
6. Updated tests accordingly.


meirroth:
Resolves #955

meirroth:
What does this error mean? [`error  Missing semicolon  semi`](https://github.com/disease-sh/API/pull/955/checks?check_run_id=1527747981#step:4:10)

meirroth:
> 
> 
> What does this error mean? [`error Missing semicolon semi`](https://github.com/disease-sh/API/pull/955/checks?check_run_id=1527747981#step:4:10)

Never mind I think I fixed it.

ebwinters (approved):

PR #965, opened 2021-01-02
Minor correction about Taiwan

Yug34:


ebwinters:
Why is this necessary? Some of these alternate names are used elsewhere in the application

Yug34:
It is not really necessary, this PR is more or less purely political. (Sorry)

Taiwan itself does not identify itself as a part of China, it is only China that is forcing overseas censorship and not agreeing to identify Taiwan as an indepentent country. Why play a part in that?

Taiwan is an excellent example of a modern liberal democracy that contrasts sharply with PRC's (Mainland China's) totalitarian state. This commitment to freedom threatens PRC so deeply that they are willing to go to war to prevent Taiwan from declaring independence.

ebwinters:
I understand the politics, but I am worried removing this would be a breaking change. There was a point where we had another source calling Taiwain "Taiwan, Province of China", so removing this alias might break code for someone trying to get to the data using this name.

PR #975, opened 2021-01-16
Smart tests

ebwinters:
# This is some cool shit
## What is this
- Basically only running the bare minimum tests for each PR
## How do I do it
1. There is a new workflow for PRs which is required that builds a specific dockerfile which does a few things
    a. Get the changed files
    b. Pipe those mfs into a script `associateTests.js` which will associate the files to tests, and write those to a file called test_to_run.txt
    c. Docker compose test command will run a script `runtests.sh` which takes these file contents as input, and provides them to mocha
2. The rest of the changes should be self explanatory. The node script associates files to tests, and the bash script is just an easy way to embed a script into a docker-compose file


EliteDaMyth (approved):
big cool

Issue #978, opened 2021-01-18
Feature Request - Change how Vaccine data is returned


lukehecht:
The daily vaccine data is returned with the date as the key (see below)
![image](https://user-images.githubusercontent.com/17548106/104954745-ba5a3900-59c0-11eb-83df-2c454c0b24b4.jpeg)
 but using numbers and special characters as JSON keys makes it a nightmare to work with. 

May I suggest reformatting the response to give each set of data its own object within the main data object? I’ve included a screenshot of the UK governments API to demonstrate 
![image](https://user-images.githubusercontent.com/17548106/104954769-c9d98200-59c0-11eb-9743-2e80e85067c4.jpeg)

chcald:
Hello, I've fixed the problem, please let me access to commit

ebwinters:
@chcald make a pull request

chcald:
I do not have permission to push, I would need to push my branch to create the PR. Thanks

ebwinters:
@chcald make a fork and push your changes there, then make a PR 😃

ebwinters:
@chcald bro?

comster:
Hmm, idk if this is the right thread but...

It would also be nice to include more data/fields which appears to be in the source like:

`total_vaccinations,people_vaccinated,people_fully_vaccinated,daily_vaccinations_raw,daily_vaccinations,total_vaccinations_per_hundred,people_vaccinated_per_hundred,people_fully_vaccinated_per_hundred,daily_vaccinations_per_million`

Total is nice, but since some require 2 doses to be "fully" vaccinated, it would be nice to have numbers that could be compared to the size of the population.

comster:
@nik-suri I see that #993 somewhat addresses this, but unfortunately I see that it's still missing useful data provided, namely `people_fully_vaccinated`.  Any chance of getting that included in the `fullData` response?

nik-suri:
@comster I see, this would require updates to the scraper `getVaccine.js` to include those data points in Redis as well. Since this issue is closed with the 2 PRs above, you may want to open the scraper enhancement as a new issue.

puf17640:
can we close this?

nik-suri:
@puf17640 Yes, this should be closed since the specific issue - reformatting the vaccine data response - has been merged in.

@comster Can you please open a new issue requesting the additional vaccine data fields?

PR #993, opened 2021-04-21
Feature request #978

nik-suri:
Addresses #978

- Updates the following endpoints to accept a new query parameter `fullData`, which is a boolean indicating whether to return the full vaccine data as a list of objects while remaining backwards-compatible:
  - `/v3/covid-19/vaccine/coverage`
  - `/v3/covid-19/vaccine/coverage/countries`
  - `/v3/covid-19/vaccine/coverage/countries/:country`
  - `/v3/covid-19/vaccine/coverage/states`
  - `/v3/covid-19/vaccine/coverage/states/:query`
- Updates tests and swagger definitions.

### DIFF
--- a/public/apidocs/swagger_v3.json
+++ b/public/apidocs/swagger_v3.json
@@ -1237,7 +1237,7 @@
           "200": {
             "description": "Status Ok",
             "schema": {
-              "$ref": "#/definitions/vaccineStateCoverage"
+              "$ref": "#/definitions/vaccineStatesCoverage"
             }
           }
         }
@@ -2229,55 +2229,48 @@
        }
     },
     "vaccineCountriesCoverage":{
-       "type":"array",
-       "items":{
-         "type":"object",
-         "properties":{
-           "country":{
-             "type":"string"
-           },
-           "timeline":{
-             "type":"object",
-             "properties":{
-               "date":{
-                 "type":"number"
-               }
-             }
-           }
-         }
-       }
+      "type":"array",
+      "items":{
+        "$ref": "#/definitions/vaccineCountryCoverage"
+      }
     },
     "vaccineCountryCoverage":{
-       "type":"object",
-       "properties":{
-         "country":{
-           "type":"string"
-         },
-         "timeline":{
-           "type":"object",
-           "properties":{
-             "date":{
-               "type":"number"
-             }
-           }
-         }
-       }
+      "type":"object",
+      "properties":{
+        "country":{
+          "type":"string"
+        },
+        "timeline":{
+          "type":"object",
+          "properties":{
+            "date":{
+              "type":"number"
+            }
+          }
+        }
+      }
+    },
+    "vaccineStatesCoverage": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/vaccineStateCoverage"
+      }
     },
     "vaccineStateCoverage":{
-       "type":"object",
-       "properties":{
-         "state":{
-           "type":"string"
-         },
-         "timeline":{
-           "type":"object",
-           "properties":{
-             "date":{
-               "type":"number"
-             }
-           }
-         }
-       }
+      "type":"object",
+      "properties":{
+        "state":{
+          "type":"string"
+        },
+        "timeline":{
+          "type":"object",
+          "properties":{
+            "date":{
+              "type":"number"
+            }
+          }
+        }
+      }
     },
     "therapeutics": {
       "properties": {
--- a/routes/v3/covid-19/apiVaccine.js
+++ b/routes/v3/covid-19/apiVaccine.js
@@ -16,13 +16,9 @@ router.get('/v3/covid-19/vaccine/coverage', async (req, res) => {
 	const data = JSON.parse(await redis.get(keys.vaccine_coverage));
 	if (data) {
 		const { world } = data;
-		const { lastdays } = req.query;
-		const numericLastDays = getLastDays(lastdays);
-		const extractedData = {};
-		Object.keys(world).slice(numericLastDays * -1).forEach((date) => {
-			extractedData[date] = world[date].total;
-		});
-		res.send(extractedData);
+		const { lastDays, isFullData } = parseVaccineQueryParams(req.query);
+		const coverageData = buildVaccineTimeline(world, lastDays, isFullData);
+		res.send(coverageData);
 	} else {
 		res.status(404).send({ message: `Error fetching vaccine coverage data` });
 	}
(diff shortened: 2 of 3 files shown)


ebwinters:
Will look this over and address your questions today @nik-suri. Thanks for contributing!

ebwinters:
Before reviewing your code, I can address your question below. I think it would be a good idea to have some query parameter that you could give like `?timelineView=true`, which would allow for re-formatting of the data for https://disease.sh/v3/covid-19/vaccine/coverage. It might also be a good idea to pass this query param to all other controllers as well, so we are not returning double the amount of data per call. It will also make it easier to deprecate the old and mal-formatted logic down the road.

public/apidocs/swagger_v3.json
```diff
@@ -1237,7 +1237,7 @@
           "200": {
             "description": "Status Ok",
             "schema": {
-              "$ref": "#/definitions/vaccineStateCoverage"
+              "$ref": "#/definitions/vaccineStatesCoverage"
```
ebwinters, line 1240: good practice 👍

nik-suri:
> Before reviewing your code, I can address your question. I think it would be a good idea to have some query parameter that you could give like `?timelineView=true`, which would allow for re-formatting of the data for https://disease.sh/v3/covid-19/vaccine/coverage. It might also be a good idea to pass this query param to all other controllers as well, so we are not returning double the amount of data per call. It will also make it easier to deprecate the old and mal-formatted logic down the road.

Sure - so passing `?timelineView=true` would return the newly formatted data?
Would `timelineView` be the best name for this property? Perhaps something like `timelineList` or `fullData` might be more descriptive (as this change will return the full data set, instead of just the totals).

routes/v3/covid-19/apiVaccine.js
```diff
@@ -36,10 +36,11 @@ router.get('/v3/covid-19/vaccine/coverage/countries', async (req, res) => {
...
 		const extractedData = countries.map((country) => {
 			const { timeline } = country;
-			const obj = { country: country.country, timeline: {} };
-			Object.keys(timeline).slice(numericLastDays * -1).forEach((date) => {
-				obj.timeline[date] = timeline[date].total;
-			});
+			const obj = {
+				country: country.country,
+				timeline: buildVaccineTimeline(timeline, numericLastDays),
+				timelineList: buildVaccineTimelineList(timeline, numericLastDays)
```
ebwinters, line 42: as my comment says, I think the `timelineList` field would be better used as a query parameter. So in this case, you could just conditionally set `timeline` based on parameter:
```js
{
    country: country.country,
    timeline: queryParam ? buildVaccineTimeline(timeline, numericLastDays) : buildVaccineTimelineList(timeline, numericLastDays)
}
```

routes/v3/covid-19/apiVaccine.js
```diff
@@ -105,18 +108,34 @@ router.get('/v3/covid-19/vaccine/coverage/states/:query', async (req, res) => {
...
+
+function buildVaccineTimeline(fullTimeline, numDays) {
+	const truncatedTimeline = {};
+	Object.keys(fullTimeline).slice(numDays * -1).forEach(date => {
+		truncatedTimeline[date] = fullTimeline[date].total;
+	});
+	return truncatedTimeline;
+}
+
+function buildVaccineTimelineList(fullTimeline, numDays) {
```
ebwinters, line 133: can you rename `buildVaccineTimelineList` to something more descriptive?

ebwinters, line 133: Can you add a docstring comment for this method to explain the difference between this and the above method

tests/v3/covid-19/api_vaccine.spec.js
```diff
@@ -81,6 +81,7 @@ describe('Testing /v3/covid-19/vaccine/ vaccine coverage', () => {
 				res.body.length.should.not.equal(0);
 				res.body[0].should.have.property('country');
 				res.body[0].should.have.property('timeline');
+				res.body[0].should.have.property('timelineList');
```
ebwinters, line 84: When you make this a query parameter you will need to add new tests using the query parameter and checking the format is correct

ebwinters (changes requested):
very simple implementation and clean code. A few comments to address though, nice job overall

routes/v3/covid-19/apiVaccine.js
nik-suri: We can make this change, however this gives rise to a new problem (of documentation).
To describe different responses in SwaggerUI, we will likely have to use `oneOf` - https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/.
Unfortunately swagger 2.0 does not support this. We would need to change the schema to openapi 3.0.0. If we change the schema, then we will have to rewrite quite a bit of the swagger_v3.json file:
- Change `#/definitions` to `#/components/schemas`
- Change each of these keys: 
```
        "responses": {
          "200": {
            "description": "Status Ok",
            "schema": {
              "$ref": "#/components/schemas/vaccineCoverage"
            }
          }
        }
```
to 
```
        "responses": {
          "200": {
            "description": "Status Ok",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/vaccineCoverage"
                }
              }
```

Would like to hear your thoughts on this - do you see an alternative solution which is simpler? Perhaps we should consider opening a new api version for the newly formatted data instead.

ebwinters:
> > Before reviewing your code, I can address your question. I think it would be a good idea to have some query parameter that you could give like `?timelineView=true`, which would allow for re-formatting of the data for https://disease.sh/v3/covid-19/vaccine/coverage. It might also be a good idea to pass this query param to all other controllers as well, so we are not returning double the amount of data per call. It will also make it easier to deprecate the old and mal-formatted logic down the road.
> 
> Sure - so passing `?timelineView=true` would return the newly formatted data?
> Would `timelineView` be the best name for this property? Perhaps something like `timelineList` or `fullData` might be more descriptive (as this change will return the full data set, instead of just the totals).

I like `fullData=true` better, yeah. Also, I didn't think about what you said regarding the swagger docs. If you're up for updating the documentation to V3, then I am totally for it. It's understandable if it's out of scope for this though, and I would also be fine returning `timeline` and `fullData` fields for a single response. It is not that much extra data anyways, so the response time won't be affected.

nik-suri:
> I like `fullData=true` better, yeah. Also, I didn't think about what you said regarding the swagger docs. If you're up for updating the documentation to V3, then I am totally for it. It's understandable if it's out of scope for this though, and I would also be fine returning `timeline` and `fullData` fields for a single response. It is not that much extra data anyways, so the response time won't be affected.

I am fine with updating the swagger docs to openapi 3.0.0. The only thing is that would be a very large change to review.
If it sounds good to you, we can split this into two PRs - this one can just add the tests/functionality and the next one can update the swagger docs.

ebwinters:
> > I like `fullData=true` better, yeah. Also, I didn't think about what you said regarding the swagger docs. If you're up for updating the documentation to V3, then I am totally for it. It's understandable if it's out of scope for this though, and I would also be fine returning `timeline` and `fullData` fields for a single response. It is not that much extra data anyways, so the response time won't be affected.
> 
> I am fine with updating the swagger docs to openapi 3.0.0. The only thing is that would be a very large change to review.
> If it sounds good to you, we can split this into two PRs - this one can just add the tests/functionality and the next one can update the swagger docs.

you can do it in separate PRs, sure. Thanks!

ebwinters:
@nik-suri are you going to make the docs PR? Don't want to merge this without it

ebwinters (approved):

nik-suri:
> @nik-suri are you going to make the docs PR? Don't want to merge this without it

@ebwinters sure, I'll open it now. I was originally planning on opening that one after this one was merged.

comster:
Sweet!  Looking forward to getting the extra data here 👍

Issue #1003, opened 2021-05-15
[BUG] https://disease.sh/v3/covid-19/continents/oceania doesn't work


nnguyen52:
Can you fix this endpoint (https://disease.sh/v3/covid-19/continents/oceania ) 

Thank you!  <3

puf17640:
I'm currently looking into it but please bear with me, I'm hungover and my laptop has 4% battery left.

puf17640:
hm, seems we never wrote tests to check if all continent endpoints are actually working... I'll post the issue in the discord server, maybe someone has time to fix this properly!

neo7337:
For this issue, it is found that here strict checking is not working properly
https://github.com/disease-sh/API/blob/master/utils/nameUtils.js#L87

in case of strict `checking==true` we should be doing fuzzySearch() otherwise we can go with the logic present here : `stringUtils.wordsStandardize(ctry[selector]) === standardizedName`

lemme know if this feels right?

in case of australia/oceania strict checking should be always true otherwise it won't get the data

neo7337:
if `strict` flag is set to `false` it works fine for oceania or australia as well.

nnguyen52:
yes @neo7337, strict = false make this endpoint works. Thank you kindly! < 3

ebwinters:
real issue is here: https://github.com/disease-sh/API/pull/1004#issuecomment-842295774

puf17640:
this has been fixed by the latest PR and some commits I pushed afterwards to extend the fix.

https://disease.sh/v3/covid-19/continents/australia-oceania,
https://disease.sh/v3/covid-19/continents/australia and
https://disease.sh/v3/covid-19/continents/oceania
should all work now and return the same data.

PR #1004, opened 2021-05-16
#1003: added test cases for /v3/covid-19/continents with /oceania or /australia

neo7337:
Test cases added for australia and oceania

### DIFF
--- a/tests/v3/covid-19/api_worldometers.spec.js
+++ b/tests/v3/covid-19/api_worldometers.spec.js
@@ -123,7 +123,7 @@ describe('TESTING /v3/covid-19/continents', () => {
 		chai.request(app)
 			.get('/v3/covid-19/continents/oceania?strict=false')
 			.end((err, res) => {
-				testBasicProperties(err, res, 200, 'object')
+				testBasicProperties(err, res, 200, 'object');
 				done();
 			});
 	});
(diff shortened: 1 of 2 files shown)


ebwinters:
does adding test cases fix the bug?

neo7337:
I don't think it's a bug, just the test cases were missing for this scenario

ebwinters:
this is the bug https://disease.sh/v3/covid-19/continents/oceania

That does not work

ebwinters:
Also why aren't the test cases running

neo7337:
> this is the bug https://disease.sh/v3/covid-19/continents/oceania
> 
> That does not work

the endpoint works with strict=false as the endpoint has strict=true set by default

https://disease.sh/v3/covid-19/continents/oceania?strict=false

neo7337:
> Also why aren't the test cases running

seems like the builds are queued

ebwinters:
https://twitter.com/githubstatus/status/1393951720946946048?s=20 lol

ebwinters:
@neo7337 here is the real issue:
"[{\"updated\":1621255355887,\"cases\":65926,\"todayCases\":10,\"deaths\":1225,\"todayDeaths\":null,\"recovered\":61834,\"todayRecovered\":6,\"active\":2867,\"critical\":3,\"casesPerOneMillion\":1542.95,\"deathsPerOneMillion\":28.67,\"tests\":19909293,\"testsPerOneMillion\":465961.46,\"population\":42727338,\"continent\":\"**Australia/Oceania**\",\"undefined\":null,\"activePerOneMillion\":67.1,\"...

We need to create some alias for Australia and Oceania to map to this key.

puf17640:
> We need to create some alias for Australia and Oceania to map to this key.

or we could split it at "/" and then just check if the specified keyword is either one of them i guess? what do you think? feels like a hotfix but i can't think of a good workaround for this

puf17640:
currently testing this PR

ebwinters (approved):

PR #1010, opened 2021-05-21
Fix/gov austria

ebwinters:
remove scraper it's broken


puf17640:
yeah they totally changed the website, i hate it

Issue #1015, opened 2021-06-10
[BUG] OAS validation errors in swagger_v3.json


tmkasun:
**Describe the bug**

The [swagger_v3.json](https://github.com/disease-sh/API/blob/master/public/apidocs/swagger_v3.json) contains about 90 errors when it's validated against OAS 3 spec.

**To Reproduce**
Where is the bug occurring? Examples:

- [Live URL in swaggerhub](https://app.swaggerhub.com/apis/tmkasun/disease-sh_docs_an_open_api_for_disease_related_statistics/3.0.0)
- Server-side? No in API definition
- Development environment? N/A

**Expected behavior**

Expect to have a few errors or as no validation errors as possible.

[Here](https://github.com/aneeshafedo/open-api-connectors/blob/master/generated-connectors/novel-covid19-modified/open-api.yaml) is a fixed version of the same API definition (bit outdated)

**Screenshots**

![image](https://user-images.githubusercontent.com/3313885/121477406-6db1e180-c9e5-11eb-8d1b-1376d86971bd.png)

PrincessRavy:
hey @aneeshafedo, want to help out with this?

mymindishazel:
that's not my PR but at this point feel free to close as stale

pujux:
oh damn, I didn't want to mention you but @tmkasun 😆 sorry there! I'll still close as stale until any further communication.

tmkasun:
I didn't open a PR ?

PR #1029, opened 2021-10-24
update OpenAPI v3 definition

DiogenesPolanco:
image


PrincessRavy:
Hi! The current OpenAPI definition is in public/apidocs/swagger_v3.json - please commit a diff to that instead of adding a file that does the same :)
Also refer to #1015

PR #1035, opened 2022-02-04
Ebwinters/fix docker pub

ebwinters:


ebwinters:
@bcmendoza I pushed new changes, is that what you mean. I am honestly pretty foreign to the concept of tagging containers and how flux works, but this should publish
- an image with `latest` tag
- an image with `sha-commithash` tag
- an image with calver tag

bcmendoza (approved):
Looks great AFAIK! Thanks for this.

ebwinters:
let's see what happens lol

